PR #3, opened 2020-02-26
ovirt dynamic plugin improvements

bverschueren:
##### SUMMARY
Improve oVirt dynamic inventory plugin based on comments in https://github.com/ansible/ansible/pull/49085

##### ISSUE TYPE
- Feature Pull Request

PR #7, opened 2020-03-09
Use fully qualified name in inventory plugin

AlanCoding:
This is a common issue that several collections hit.

Basically, if you don't use the fully qualified name here, there's no way to use this inventory because Ansible core expects it under the collection name.namespace.

I can't yet say that I've tested this, and I don't think I'll get full validation, but I'll try to check back after I manually test and make sure that with an invalid server, it gives the error I would expect.


ovirt-infra:

mnecas:
@AlanCoding thank you for the report.

PR #9, opened 2020-03-10
fix big migration

mnecas:
After the "Big Migration" there was missing some small changes https://github.com/oVirt/ovirt-ansible-collection/issues/8


mnecas:
I have added CloudRetry so the rpm is independent and when community collection have rpm we can update the path to the plugin.

Issue #14, opened 2020-03-24
Rename collection?


gundalow:
Is there a name the collection is called `ovirt.ovirt_collection` rather than something shorter?

As a reminder the collection name doesn't need to match the GitHub repo name

mnecas:
Is there some kind of standard for naming the collections?
If there is no particular reason why we should change it I would keep it because we already released it to the galaxy.

WDYT @mwperina

mwperina:
Well, we thought about it, but `ovirt.ovirt` doesn't make sense at all and just `ovirt` cannot be used.

I've looked at names of other collection names and there is often used `.cloud` suffix (openstack, google, ...), so maybe `ovirt.cloud` would be better, what do you think? Are there any best practices around collection naming, John?

@gundalow

mnecas:
I'm not sure how it will be in automation hub if we will have collection under some different namespace than the name `ovirt.ovirt` would not be worst.
Do you have some information about the automation hub @gundalow ?

abenokraitis:
`ovirt.ovirt` would be consistent with other naming of upstream collections, and would be simpler than putting the word `collection` in there, since that's redundant (project_name.collection_name.asset). The downstream Red Hat supported collection in Automation hub can have the same collection name but must be part of the `redhat.*` namespace.

mnecas:
Got two questions:
- If we rename the collection can we somehow remove the old collection from the galaxy or if we could update somehow the existing name?
- When we will use namespace redhat we will need to adapt whole collection to it?

anshulbehl:
@mnecas for the answer to first question:
You cannot remove or update the existing name of a collection, they will be uploaded as new collection. The only option is either to 
- Deprecate the old collection and upload the new one as `ovirt.ovirt`
- You can create the new collection and keep the old one as well, but the new version of old collection can be empty and will only have new one as a dependency which means whenever a user install the old one, new one will come with it.


for the answer to your second question:
Relative imports works inside collections, so you can do away with absolute imports that you have used like below (example)
`ansible_collections.ovirt.ovirt_collection.plugins.module_utils.ovirt import check_sdk`

and replace it with
`from ..module_utils.ovirt import check_sdk`

This way you can avoid namespace collisions  in your src when maintaining for two namespaces.
However, keep in mind that relative imports don't work for doc_fragments and you will have to change then with namespace.
[...]

PR #16, opened 2020-04-07
Rename collection

mnecas:
After a request from the ansible core team https://github.com/oVirt/ovirt-ansible-collection/issues/14. We have decided to rename `ovirt.ovirt_collection` to `ovirt.ovirt`.

cc @mwperina

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,14 @@
 oVirt ansible collection
 ====================================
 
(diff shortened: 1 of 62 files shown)


mnecas:
Added branch ovirt_collection if we need to do some updates on the deprecated collection.

PR #17, opened 2020-04-08
Add migration_encrypted option

mnecas:
Add migration_encrypted option.
After https://github.com/oVirt/ovirt-ansible-collection/pull/13 we can migrate original PR https://github.com/ansible/ansible/pull/65916.
@dangel101 @mwperina

### DIFF
--- a/plugins/modules/ovirt_cluster.py
+++ b/plugins/modules/ovirt_cluster.py
@@ -179,6 +179,14 @@ options:
             - "C(false) - Override the global setting to I(false)."
             - "C(inherit) - Use value which is set globally."
         choices: ['true', 'false', 'inherit']
+    migration_encrypted:
+        description:
+            - "If I(True) encryption is used during live migration of the virtual machine."
+            - "Following options are supported:"
+            - "C(true) - Override the global setting to I(true)."
+            - "C(false) - Override the global setting to I(false)."
+            - "C(inherit) - Use value which is set globally."
+        choices: ['true', 'false', 'inherit']
     migration_policy:
         description:
             - "A migration policy defines the conditions for live migrating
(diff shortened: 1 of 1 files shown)


plugins/modules/ovirt_cluster.py
```diff
@@ -179,6 +179,14 @@
             - "C(false) - Override the global setting to I(false)."
             - "C(inherit) - Use value which is set globally."
         choices: ['true', 'false', 'inherit']
+    migration_encrypted:
+        description:
+            - "If I(True) encryption is used during live migration of the virtual machine."
+            - "Following options are supported:"
+            - "C(true) - Override the global setting to I(true)."
```
mwperina, line 186: Shouldn't we use `True` instead of `true`?

plugins/modules/ovirt_cluster.py
```diff
@@ -179,6 +179,14 @@
...
+            - "C(false) - Override the global setting to I(false)."
```
mwperina, line 187: Shouldn't we use `False` instead of `false`?

plugins/modules/ovirt_cluster.py
```diff
@@ -179,6 +179,14 @@
...
+            - "C(inherit) - Use value which is set globally."
+        choices: ['true', 'false', 'inherit']
```
mwperina, line 189: `choices: ['True', 'False', 'inherit']`

plugins/modules/ovirt_cluster.py
mnecas: The smaller case is used in `migration_compressed` and `migration_auto_converge` so it is better to keep it same everywhere.

mwperina: Shouldn't we alway use the values True/False which are compatible with Python boolean values?

mnecas: The type of the `migration_encrypted` is a string and the ovirt sdk can take the value with first letter small or large case, so it would not affect anything. 
I want to keep it as it is because it is same implementation as in `migration_compressed` and `migration_auto_converge`. 
If you think we should update it to large I'll update it everywhere.

mwperina: OK

PR #19, opened 2020-04-14
ovirt_disk: upload_image_path autodetect size

mnecas:
This is a feature when user does not specify disk size when uploading the disk. 
It will be automatically determined for him.

@mwperina  @dangel101


knumskull:
Great work. I've tested this in my environment and it worked like a charm and removed the error-prone size calculation. Now it's much more feasible to use this solution in an automated way, when you don't know the initial iso size.

Issue #20, opened 2020-04-14
Can't delete a vlan tag assigned to a network 


elroncio:
If I create a VLAN Tag for a network I can't delete it, but I have to delete the network and recreate it without the tag. 
If I use the web console I can easily remove a vlan tag on a network.

I'm currently using Ansible 2.9.6.

PR #21, opened 2020-04-14
ovirt_network: add support of removing vlan_tag

mnecas:
Usually, to remove some parameter we specify the variable as an empty string. 
The vlan_tag is specified as int so I have added option to remove the VLAN tag when you use -1. 

Because this is not RFE will try to push it to ansible 2.9.z.

Fixes https://github.com/oVirt/ovirt-ansible-collection/issues/20

@mwperina @dangel101

### DIFF
--- a/plugins/modules/ovirt_network.py
+++ b/plugins/modules/ovirt_network.py
@@ -59,6 +59,8 @@ options:
     vlan_tag:
         description:
             - "Specify VLAN tag."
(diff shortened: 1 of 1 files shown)


elroncio:
@mnecas, sorry to bother you, but is this already ported to Ansible 2.9?

mnecas:
@elroncio oh sorry, forgot about the backport, will do it, thanks for the ping

PR #22, opened 2020-04-15
ovirt_disk: add upload image warning for correct format

mnecas:
Add warning when uploading iso and the format is cow, before this PR the upload would end on FINALIZING_FAILURE and no more information, now it shows warning which will inform the user to update to raw.

@mwperina  @dangel101

### DIFF
--- a/plugins/modules/ovirt_disk.py
+++ b/plugins/modules/ovirt_disk.py
@@ -734,7 +734,7 @@ def main():
             # Upload disk image in case it's new disk or force parameter is passed:
             if module.params['upload_image_path'] and (is_new_disk or module.params['force']):
                 if module.params['format'] == 'cow' and module.params['content_type'] == 'iso':
-                    module.warn("Parameter 'format' should not be 'cow' when uploading iso image, please change it to 'raw'.")
+                    module.warn("To upload an ISO image 'format' parameter needs to be set to 'raw'.")
                 uploaded = upload_disk_image(connection, module)
                 ret['changed'] = ret['changed'] or uploaded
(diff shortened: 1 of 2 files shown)


plugins/modules/ovirt_disk.py
```diff
@@ -731,6 +733,8 @@ def main():
 
             # Upload disk image in case it's new disk or force parameter is passed:
             if module.params['upload_image_path'] and (is_new_disk or module.params['force']):
+                if module.params['format'] == 'cow' and module.params['content_type'] == 'iso':
+                    module.warn("Parameter 'format' should not be 'cow' when uploading iso image, please change it to 'raw'.")
```
mwperina, line 737: Please rephrase: `To upload an ISO image 'format' parameter needs to be set to 'raw'`

plugins/modules/ovirt_disk.py
```diff
@@ -73,6 +74,7 @@
             - Specify format of the disk.
             - Note that this option isn't idempotent as it's not currently possible to change format of the disk via API.
         choices: ['raw', 'cow']
+        default: 'cow'
```
mwperina, line 77: Does it make sense to have a default format?

mnecas: Not sure.
There is already a default format [1], this just updates the documentation.
[1] https://github.com/oVirt/ovirt-ansible-collection/blob/master/plugins/modules/ovirt_disk.py#L654

mwperina: OK, I missed it's already default in the code

PR #30, opened 2020-04-17
Clarify cloud-init custom_script documentation

snecklifter:
It isn't immediately clear what can go in the custom_script field
so the additional line of text will point users in the right
direction. This commit adds this. We could also add a link but
am personally against external URLs as they represent moving
targets.


snecklifter:
I had to hunt around to find the answer at https://cloudinit.readthedocs.io/en/latest/topics/modules.html so hopefully the above commit will help.

PR #33, opened 2020-04-24
ovirt_host_network: fix save of network config

mnecas:
This is a workaround to allow save until the fixed in the engine.
Proper fixed should be in https://bugzilla.redhat.com/show_bug.cgi?id=1798818
https://github.com/ansible/ansible/issues/67810
@dangel101 @mwperina

### DIFF
--- a/plugins/modules/ovirt_host_network.py
+++ b/plugins/modules/ovirt_host_network.py
@@ -579,7 +579,7 @@ def main():
                     ] if labels else None,
                     removed_network_attachments=attachments if attachments else None,
                 )
-                if engine_supported(connection, '4.3'):
+                if engine_supported(connection, '4.4'):
                     setup_params['commit_on_success'] = module.params['save']
                 elif module.params['save']:
                     setup_params['post_action'] = host_networks_module._action_save_configuration


plugins/modules/ovirt_host_network.py
```diff
@@ -579,7 +579,7 @@ def main():
                     ] if labels else None,
                     removed_network_attachments=attachments if attachments else None,
                 )
-                if engine_supported(connection, '4.3'):
+                if engine_supported(connection, '4.4'):
```
mwperina, line 582: Please add a comment: ```# TODO: Remove after https://bugzilla.redhat.com/1798818 is fixed

mnecas: I would keep it as it is if there was issue in 4.3 and will get fixed in 4.4. It would be better to use the older method of the save configuration `host_networks_module._action_save_configuration`.

mwperina: OK

mnecas:
ci test please

mwperina (approved):
LGTM

plugins/modules/ovirt_host_network.py
mwperina: https://bugzilla.redhat.com/show_bug.cgi?id=1798818 is included in oVirt 4.4.1, let's merge it

PR #34, opened 2020-04-28
ovirt inventory add creation_time

mnecas:
add `creation_time` to ovirt inventory

### DIFF
--- a/plugins/inventory/ovirt.py
+++ b/plugins/inventory/ovirt.py
@@ -127,7 +127,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             'fqdn': vm.fqdn,
             'os_type': vm.os.type,
             'template': self.connection.follow_link(vm.template).name,
-            'creation_time': vm.creation_time,
+            'creation_time': str(vm.creation_time),
             'tags': [tag.name for tag in tags],
             'affinity_labels': [label.name for label in labels],
             'affinity_groups': [
--- a/plugins/inventory/ovirt.py
+++ b/plugins/inventory/ovirt.py
@@ -127,6 +127,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             'fqdn': vm.fqdn,
             'os_type': vm.os.type,
             'template': self.connection.follow_link(vm.template).name,
+            'creation_time': vm.creation_time,
             'tags': [tag.name for tag in tags],
             'affinity_labels': [label.name for label in labels],
             'affinity_groups': [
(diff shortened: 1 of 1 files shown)


plugins/inventory/ovirt.py
```diff
@@ -127,6 +128,8 @@ def _get_dict_of_struct(self, vm):
             'fqdn': vm.fqdn,
             'os_type': vm.os.type,
             'template': self.connection.follow_link(vm.template).name,
+            'creation_time': str(vm.creation_time),
+            'creation_time_timestamp': vm.creation_tim.timestamp() if vm.creation_tim.tzinfo else vm.creation_tim.replace(tzinfo=timezone.utc).timestamp(),
```
kedark3, line 132: Following would work:

``` 
vm.creation_time.strftime('%s') # For Python2
vm.creation_time.timestamp() # For python3
```
And my Ansible Tower is python2 default as its RHEL 7 based. For RHEL8+Py3 based `.timestamp()` would work. Its now up to you if you want to try supporting Py2 which is dead. I guess its my team who need to update RHEL8 if we want to make this work for us.

kedark3: If you do NOT want to care for Py2, then update your PR as follows please:
```
            'creation_time': str(vm.creation_time),
            'creation_time_timestamp': vm.creation_time.timestamp()
```
That's all, nothing else, you can remove if/else and importing timezone. Thanks for opening this PR.

mnecas: This script is created for python3 `#!/usr/bin/python3` and we distribute the collection rpm to el8 and fc30.
I'll remove the if statement and leave it as it is.

AlanCoding: > This script is created for python3 #!/usr/bin/python3 and we distribute the collection rpm to el8 and fc30.

The shebang has no effect as I understand it, because this is not called in a subprocess like the old scripts were. It is imported directly into the main Ansible process in the inventory loading stage.

You can confirm this by running in a python2 virtual environment and applying:

```diff
diff --git a/plugins/inventory/ovirt.py b/plugins/inventory/ovirt.py
index 3a8736f..73c88a4 100755
--- a/plugins/inventory/ovirt.py
+++ b/plugins/inventory/ovirt.py
@@ -106,6 +106,8 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
              :return dict of vm struct type
         '''
 
+        import sys
+        raise Exception('version: ' + str(sys.version))
         vms_service = self.connection.system_service().vms_service()
         clusters_service = self.connection.system_service().clusters_service()
         vm_service = vms_service.vm_service(vm.id)
```

If you're invoking Ansible itself with python2, this runs with python2.

Dropping support for python2 isn't crazy, of course, but if you do, you should put it in the requirements for the plugin, because it's not implied.

PR #35, opened 2020-05-02
Support direct upload/download to Image I/O daemon on oVirt node

### LINKED ISSUE #29: ovirt_disk should use new code to upload image [COPY]
This is a copy of the ansible/ansible issue.
The original issue: https://github.com/ansible/ansible/issues/68403

ganto:
Fixes #29 and therefore fixes image down-/upload on oVirt 4.4 engines also see [RHBZ 1801710](https://bugzilla.redhat.com/show_bug.cgi?id=1801710).

I verified this code with the following testing playbook:
```
---                                                                                                                                                                                                                                                                                                                   
- hosts: localhost
  connection: local                   
  gather_facts: False
                                                                               
  collections:
    - ovirt.ovirt

  tasks:
    - ovirt_auth:
        url: https://ovirt.example.com/ovirt-engine/api
        username: admin@internal
        password: password

    - ovirt_disk:
        auth: '{{ ovirt_auth }}'
        name: testdisk
        size: 10GiB
        format: cow
        upload_image_path: /tmp/testdisk.img
        storage_domain: vmstore

    - ovirt_disk:
        auth: '{{ ovirt_auth }}'
        name: testdisk
        download_image_path: /tmp/testdisk-1.img
        storage_domain: vmstore

    - ovirt_auth:
        ovirt_auth: '{{ ovirt_auth }}'
        state: absent
```
Before this patch this playbook would fail with the following error when run against a oVirt 4.4 API:
```
TASK [ovirt_disk] ******************************************************************************************************************************************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ConnectionRefusedError: [Errno 111] Connection refused
fatal: [localhost]: FAILED! => {"changed": false, "msg": "[Errno 111] Connection refused"}
```
I also needed to [remove the authentication header](https://github.com/ganto/ovirt-ansible-collection/commit/f281406f3273a49ad9c072fd913fc5a3c9d0b1e0#diff-dfc3119277e6ad415795e92381e14ef8L421) from the download request. Otherwise I would get an error like:
```
The full traceback is:                                                                                                                                                                                                                                                                                                        
  File "/tmp/ansible_ovirt_disk_payload_wyxbh_tl/ansible_ovirt_disk_payload.zip/ansible_collections/ovirt/ovirt/plugins/modules/ovirt_disk.py", line 753, in main                                                                                                                                                             
  File "/tmp/ansible_ovirt_disk_payload_wyxbh_tl/ansible_ovirt_disk_payload.zip/ansible_collections/ovirt/ovirt/plugins/modules/ovirt_disk.py", line 454, in download_disk_image                                                                                                                                              
  File "/tmp/ansible_ovirt_disk_payload_wyxbh_tl/ansible_ovirt_disk_payload.zip/ansible_collections/ovirt/ovirt/plugins/modules/ovirt_disk.py", line 396, in transfer                                                                                                                                                         
  File "/tmp/ansible_ovirt_disk_payload_wyxbh_tl/ansible_ovirt_disk_payload.zip/ansible_collections/ovirt/ovirt/plugins/modules/ovirt_disk.py", line 435, in _transfer                                                                                                                                                        
  File "/usr/lib/python3.7/http/client.py", line 1262, in request                                                                                                                                                                                                                                                             
    self._send_request(method, url, body, headers, encode_chunked)                                                                                                                                                                                                                                                            
  File "/usr/lib/python3.7/http/client.py", line 1303, in _send_request                                                                                                                                                                                                                                                       
    self.putheader(hdr, value)                                                                                                                                                                                                                                                                                                
  File "/usr/lib/python3.7/http/client.py", line 1239, in putheader                                                                                                                                                                                                                                                           
    if _is_illegal_header_value(values[i]):                                                                                                                                                                                                                                                                                   
TypeError: expected string or bytes-like object
```
I don't don't understand what is going wrong here. Maybe someone can make some sense of this?

Unfortunately I don't have a oVirt 4.3 environment around for testing the `use_proxy` flag. ~~Also not sure if this should be enabled or disabled by default. Is there some way to autodetect this and do the right thing for 4.3 and 4.4?~~

### DIFF
--- a/plugins/modules/ovirt_disk.py
+++ b/plugins/modules/ovirt_disk.py
@@ -186,11 +186,6 @@ options:
             - I(True) if the disk should be activated.
             - When creating disk of virtual machine it is set to I(True).
         type: bool
-    use_proxy:
-        description:
-            - "Use Image I/O proxy when uploading or downloading disk image. Set
-              this to I(True) if you cannot directly connect to the oVirt node."
-        type: bool
 extends_documentation_fragment: ovirt.ovirt.ovirt
 '''
 
@@ -337,7 +332,6 @@ from ansible_collections.ovirt.ovirt.plugins.module_utils.ovirt import (
     get_dict_of_struct,
     search_by_name,
     wait,
-    engine_supported,
 )
 
 
@@ -372,10 +366,6 @@ def transfer(connection, module, direction, transfer_func):
             time.sleep(module.params['poll_interval'])
             transfer = transfer_service.get()
 
-        if module.params['use_proxy'] or not engine_supported(connection, '4.4'):
-            destination_url = urlparse(transfer.proxy_url)
-        else:
-            destination_url = urlparse(transfer.transfer_url)
         context = ssl.create_default_context()
         auth = module.params['auth']
         if auth.get('insecure'):
@@ -384,18 +374,35 @@ def transfer(connection, module, direction, transfer_func):
         elif auth.get('ca_file'):
             context.load_verify_locations(cafile=auth.get('ca_file'))
 
-        transfer_connection = HTTPSConnection(
-            destination_url.hostname,
-            destination_url.port,
-            context=context,
-        )
+        try:
+            destination_url = urlparse(transfer.transfer_url)
+            transfer_connection = HTTPSConnection(
+                destination_url.hostname,
+                destination_url.port,
+                context=context,
+            )
+            transfer_func(
+                transfer_service,
+                transfer_connection,
+                destination_url,
+                transfer.signed_ticket
+            )
+        except Exception as e:
+            module.warn("Image transfer fail with message: {}.".format(e))
+            module.warn("Retrying trasfer with proxy.")
+            destination_url = urlparse(transfer.proxy_url)
+            transfer_connection = HTTPSConnection(
+                destination_url.hostname,
+                destination_url.port,
+                context=context,
+            )
 
-        transfer_func(
-            transfer_service,
-            transfer_connection,
-            destination_url,
-            transfer.signed_ticket
-        )
+            transfer_func(
+                transfer_service,
+                transfer_connection,
+                destination_url,
+                transfer.signed_ticket
+            )
         return True
     finally:
         transfer_service.finalize()
@@ -673,7 +680,6 @@ def main():
         host=dict(default=None),
         wipe_after_delete=dict(type='bool', default=None),
         activate=dict(default=None, type='bool'),
-        use_proxy=dict(default=None, type='bool'),
     )
     module = AnsibleModule(
         argument_spec=argument_spec,
--- a/plugins/modules/ovirt_disk.py
+++ b/plugins/modules/ovirt_disk.py
@@ -355,8 +355,7 @@ def create_transfer_connection(module, transfer, context, connect_timeout=10, re
         connection.connect()
     except OSError as e:
         # Typically ConnectionRefusedError or socket.gaierror.
-        module.warn("Cannot connect to {}, trying {}: {}"
-                    .format(transfer.transfer_url, transfer.proxy_url, e))
+        module.warn("Cannot connect to %s, trying %s: %s" % (transfer.transfer_url, transfer.proxy_url, e))
 
         url = urlparse(transfer.proxy_url)
         connection = HTTPSConnection(
@@ -394,12 +393,11 @@ def transfer(connection, module, direction, transfer_func):
         elif auth.get('ca_file'):
             context.load_verify_locations(cafile=auth.get('ca_file'))
 
-        transfer_connection, url = create_transfer_connection(module, transfer, context)
+        transfer_connection, transfer_url = create_transfer_connection(module, transfer, context)
         transfer_func(
             transfer_service,
             transfer_connection,
-            url,
-            transfer.signed_ticket
+            transfer_url,
         )
         return True
     finally:
@@ -430,9 +428,9 @@ def transfer(connection, module, direction, transfer_func):
 
 
 def download_disk_image(connection, module):
-    def _transfer(transfer_service, transfer_connection, destination_url, transfer_ticket):
+    def _transfer(transfer_service, transfer_connection, transfer_url):
         BUF_SIZE = 128 * 1024
-        transfer_connection.request('GET', destination_url.path)
+        transfer_connection.request('GET', transfer_url.path)
         r = transfer_connection.getresponse()
         path = module.params["download_image_path"]
         image_size = int(r.getheader('Content-Length'))
@@ -455,12 +453,12 @@ def download_disk_image(connection, module):
 
 
 def upload_disk_image(connection, module):
-    def _transfer(transfer_service, transfer_connection, destination_url, transfer_ticket):
+    def _transfer(transfer_service, transfer_connection, transfer_url):
         BUF_SIZE = 128 * 1024
         path = module.params['upload_image_path']
 
         image_size = os.path.getsize(path)
-        transfer_connection.putrequest("PUT", destination_url.path)
+        transfer_connection.putrequest("PUT", transfer_url.path)
         transfer_connection.putheader('Content-Length', "%d" % (image_size,))
         transfer_connection.endheaders()
         with open(path, "rb") as disk:
--- a/plugins/modules/ovirt_disk.py
+++ b/plugins/modules/ovirt_disk.py
@@ -337,6 +337,7 @@ from ansible_collections.ovirt.ovirt.plugins.module_utils.ovirt import (
     get_dict_of_struct,
     search_by_name,
     wait,
+    engine_supported,
 )
 
 
@@ -371,7 +372,7 @@ def transfer(connection, module, direction, transfer_func):
             time.sleep(module.params['poll_interval'])
             transfer = transfer_service.get()
 
-        if module.params['use_proxy']:
+        if module.params['use_proxy'] or not engine_supported(connection, '4.4'):
             destination_url = urlparse(transfer.proxy_url)
         else:
             destination_url = urlparse(transfer.transfer_url)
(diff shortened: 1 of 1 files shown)


ovirt-infra:

mnecas:
ci test please

mnecas:
Thank you for the PR!
I'll jump on it and try to test it with 4.3 and 4.4 and review it.
@nirs please also check the PR.

mnecas:
when trying your changes on 4.4 I get 
```
TASK [ovirt_disk] **************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: Exception: Error occurred while uploading image. The transfer is in finalizing_failure
fatal: [10-37-137-37.rhev.lab.eng.brq.redhat.com]: FAILED! => {"changed": false, "msg": "Error occurred while uploading image. The transfer is in finalizing_failure"}
```
Will do more investigation.

plugins/modules/ovirt_disk.py
```diff
@@ -371,7 +372,7 @@ def transfer(connection, module, direction, transfer_func):
             time.sleep(module.params['poll_interval'])
             transfer = transfer_service.get()
 
-        if module.params['use_proxy']:
+        if module.params['use_proxy'] or not engine_supported(connection, '4.4'):
```
nirs, line 375: What does engine_supported(connection, '4.4') mean?

The proxy should be used only in 2 cases:
- you cannot access the host, regardless of engine version
- the user ask to use the proxy

plugins/modules/ovirt_disk.py
```diff
@@ -672,7 +673,7 @@ def main():
         host=dict(default=None),
         wipe_after_delete=dict(type='bool', default=None),
         activate=dict(default=None, type='bool'),
-        use_proxy=dict(default=False, type='bool'),
+        use_proxy=dict(default=None, type='bool'),
```
nirs, line 676: Why change to None? use_proxy should be False by default.

plugins/modules/ovirt_disk.py
```diff
@@ -186,6 +186,11 @@
             - I(True) if the disk should be activated.
             - When creating disk of virtual machine it is set to I(True).
         type: bool
+    use_proxy:
+        description:
+            - "Use Image I/O proxy when uploading or downloading disk image. Set
+              this to I(True) if you cannot directly connect to the oVirt node."
```
nirs, line 192: Looks good, but we should also warn that using the proxy is less efficient and not recommended.

In 4.4 the proxy can be 10 times faster than the proxy in 4.3, but it is still 2 times slower compared with
direct upload. These numbers are not official since scale team never test this, but it illustrates the difference.

mnecas:
Thank you for the info, I will create another PR with the imageio client.

plugins/modules/ovirt_disk.py
ganto: This guarantees backwards compatibility. As far as I understood the Ansible code against oVirt <=4.3 so far did only know about proxy mode, that's why I didn't want to change that by default.

plugins/modules/ovirt_disk.py
ganto: First I defaulted to not using a proxy which I guess would make sense for oVirt 4.4 upload directly to the node. But this would break existing oVirt 4.3 setups where the upload only works via proxy and direct node access is not possible (e.g. blocked via firewall).

plugins/modules/ovirt_disk.py
ganto: Personally I don't care how fast this upload is. The main point is that it works. I had a playbook using this module against an existing oVirt 4.3 which worked fine. Then I found it to be broken by oVirt 4.4 so I fixed the module and thought I could share it. If there is a better way to do it, go ahead with your solution.

plugins/modules/ovirt_disk.py
mnecas: I had no issues when using `use_proxy: False` on oVirt 4.3. I was able to upload qcow and raw images without a problem.

mnecas:
ci test please

plugins/modules/ovirt_disk.py
```diff
@@ -384,18 +374,35 @@ def transfer(connection, module, direction, transfer_func):
...
+            )
+        except Exception as e:
+            module.warn("Image transfer fail with message: {}.".format(e))
+            module.warn("Retrying trasfer with proxy.")
+            destination_url = urlparse(transfer.proxy_url)
+            transfer_connection = HTTPSConnection(
+                destination_url.hostname,
+                destination_url.port,
+                context=context,
+            )
```
nirs, line 398: The idea to use the proxy a fallback is good, but the implementation is not.

Lets say you upload 100 GiB image, and the uploads fails after 80 GiB, maybe because storage does not have enough space. This code will fail after sending 80 GiB, and then try to do the same from the proxy, that will fail again.

What we need instead is:

```python
    def create_connection(transfer, context, connect_timeout=10, read_timeout=60):
        url = urlparse(transfer.transfer_url)
        connection = HTTPSConnection(
            url.netloc, context=context, timeout=connect_timeout)
        
        try:
            connection.connect()
        except OSError as e:
            # Typically ConnectionRefusedError or socket.gaierror.
            log.warning(
                "Cannot connect to %s, trying %s: %s",
                transfer.transfer_url, transfer.proxy_url, e)

            url = urlparse(transfer.proxy_url)
            connection = HTTPSConnection(
                url.netloc, context=context, timeout=connect_timeout)
            connection.connect()

        connection.sock.settimeout(read_timeout)
        return connection
```

Once you created a connection, you can use it in the same way with both host and proxy.

mnecas:
ci test please

plugins/modules/ovirt_disk.py
```diff
@@ -374,35 +393,13 @@ def transfer(connection, module, direction, transfer_func):
...
-                context=context,
-            )
-
-            transfer_func(
-                transfer_service,
-                transfer_connection,
-                destination_url,
-                transfer.signed_ticket
-            )
+        transfer_connection, url = create_transfer_connection(module, transfer, context)
```
nirs, line 397: In this context, using "connection" instead of transfer_connection would be little nicer. You agree with me since
you used "url" and not "transfer_url" :-)

You can also rename url to dest_url to be more consistent with other names.

plugins/modules/ovirt_disk.py
```diff
@@ -347,6 +347,25 @@ def _search_by_lun(disks_service, lun_id):
     return res[0] if res else None
 
 
+def create_transfer_connection(module, transfer, context, connect_timeout=10, read_timeout=60):
```
nirs, line 350: It can be useful to expose the timeouts to the user so they can override the timeouts if needed.

plugins/modules/ovirt_disk.py
```diff
@@ -374,35 +393,13 @@ def transfer(connection, module, direction, transfer_func):
...
-                transfer_connection,
-                destination_url,
-                transfer.signed_ticket
-            )
+        transfer_connection, url = create_transfer_connection(module, transfer, context)
+        transfer_func(
+            transfer_service,
+            transfer_connection,
+            url,
+            transfer.signed_ticket
```
nirs, line 402: transfer.signed_ticket is not needed since 4.2 when uploading to proxy,
and we never needed when uploading to host.
See this patch explaning the issue and removing it from the SDK.
https://gerrit.ovirt.org/c/108534/

Do not add new code using transfer.signed_ticket, your code will break in 4.4.

plugins/modules/ovirt_disk.py
mnecas: cannot use `connection` it is already defined as another variable 
so if we want to be consistent maybe use the `transfer_url`

plugins/modules/ovirt_disk.py
```diff
@@ -355,7 +355,8 @@ def create_transfer_connection(module, transfer, context, connect_timeout=10, re
         connection.connect()
     except OSError as e:
         # Typically ConnectionRefusedError or socket.gaierror.
-        module.warn("Cannot connect to %s, trying %s: %s", transfer.transfer_url, transfer.proxy_url, e)
+        module.warn("Cannot connect to {}, trying {}: {}"
+                    .format(transfer.transfer_url, transfer.proxy_url, e))
```
nirs, line 359: Formatting messages for logs is bad idea. Formatting is expensive and should be done  only when the log message is actually logged. For warnings this is usually the case, but for debug message it is always not and having 2 ways to log in the same code is bad.

I hope this is not the way ansible scripts use logging. If it is I guess we want to keep this consistent with other logs here.

plugins/modules/ovirt_disk.py
mnecas: yeah thought about that but we have default timeout 3min and not sure if it would be best to wait that long

plugins/modules/ovirt_disk.py
mnecas: okay will remove it

plugins/modules/ovirt_disk.py
mnecas: This does is not for logging it just shows a user warning message. 
And around formatting, I don't mind to use it as it was before.

plugins/modules/ovirt_disk.py
```diff
@@ -355,8 +355,7 @@ def create_transfer_connection(module, transfer, context, connect_timeout=10, re
         connection.connect()
     except OSError as e:
         # Typically ConnectionRefusedError or socket.gaierror.
-        module.warn("Cannot connect to {}, trying {}: {}"
-                    .format(transfer.transfer_url, transfer.proxy_url, e))
+        module.warn("Cannot connect to %s, trying %s: %s" % (transfer.transfer_url, transfer.proxy_url, e))
```
nirs, line 358: This is the same as using {}. Please check module.warn() interface - does it work like Pyhon logging, accepting
fmt, *args or it a single srting?

Then check how other code in this area generate messages and use the same style.

plugins/modules/ovirt_disk.py
nirs: I would this to future work.

plugins/modules/ovirt_disk.py
mnecas: Oh that was your question it only accepts one string no other args. 
And it is only message the warning styling is done automatically.

PR #39, opened 2020-05-08
Use imageio client when uploading disk

### LINKED ISSUE #29: ovirt_disk should use new code to upload image [COPY]

mnecas:
Fixes #29 
Continuing of the https://github.com/oVirt/ovirt-ansible-collection/pull/35

This raises a minor issue when installing from galaxy there will be missing the `ovirt-imageio-client` python library because we cannot specify dependency of rpm.
I have moved the import to the transfer function so it would only fail when trying to upload/download.

@mwperina @dangel101 @nirs
cc @ganto

### DIFF
--- a/plugins/modules/ovirt_disk.py
+++ b/plugins/modules/ovirt_disk.py
@@ -340,7 +340,6 @@ from ansible_collections.ovirt.ovirt.plugins.module_utils.ovirt import (
     search_by_name,
     wait,
 )
-from ovirt_imageio import client
 
 
 def _search_by_lun(disks_service, lun_id):
@@ -356,6 +355,7 @@ def _search_by_lun(disks_service, lun_id):
 
 
 def transfer(connection, module, direction):
+    from ovirt_imageio import client
     transfers_service = connection.system_service().image_transfers_service()
     transfer = transfers_service.add(
         otypes.ImageTransfer(
--- a/plugins/modules/ovirt_disk.py
+++ b/plugins/modules/ovirt_disk.py
@@ -186,6 +186,11 @@ options:
             - I(True) if the disk should be activated.
             - When creating disk of virtual machine it is set to I(True).
         type: bool
+    use_proxy:
+        description:
+            - I(True) if the disk should be activated.
+            - asdasd
+        type: str
 extends_documentation_fragment: ovirt.ovirt.ovirt
 '''
 
@@ -312,6 +317,7 @@ import time
 import traceback
 import ssl
 
+from ovirt_imageio import client
 from ansible.module_utils.six.moves.http_client import HTTPSConnection, IncompleteRead
 from ansible.module_utils.six.moves.urllib.parse import urlparse
 try:
@@ -332,6 +338,7 @@ from ansible_collections.ovirt.ovirt.plugins.module_utils.ovirt import (
     get_dict_of_struct,
     search_by_name,
     wait,
+    engine_supported
 )
 
 
@@ -385,7 +392,7 @@ def transfer(connection, module, direction, transfer_func):
             transfer_service,
             proxy_connection,
             proxy_url,
-            transfer.signed_ticket
+            transfer
         )
         return True
     finally:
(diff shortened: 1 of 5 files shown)


nirs:
ovirt-imageio-client (depening on ovirt-imageio-common and qemu-img) should exist on the host
running the upload. The important dependency is qemu-img, providing /usr/bin/qemu-nbd. The upload pipeline is:

```
|           upload host        |    |            hypervisor        |   
|                              |    |                              |   
|  qemu-nbd -> imageio client  | -> |  imageio server -> qemu-nbd  |
```
When using the proxy the pipeline is:

```
|           upload host        |    |    proxy host    |    |          hypervisor          |
|                              |    |                  |    |                              |   
|  qemu-nbd -> imageio client  | -> |  imageio-server  | -> |  imageio server -> qemu-nbd  |
```

qemu-nbd is the critical part, this enables support for qcow2 images and sparseness support (e.g detecting zeroes in uploaded images). This also enable on-the-fly format conversion (e.g. upload qcow2 compressed image to raw disk).

This probably does not fit the ansible model of copying code to some machine and running it, but this is the right way to upload images.

mnecas:
If I understand it correctly and we need to have the qemu-img wouldn't it be better to install the package when doing host setup?

nirs:
qemu-img is always installed in hosts, it is used by vdsm. I think we will also add ovirt-imageio-client to the hosts. I guess it should be installed by ovirt host deploy.

plugins/modules/ovirt_disk.py
```diff
@@ -366,27 +372,27 @@ def transfer(connection, module, direction, transfer_func):
...
+            destination_url = transfer.proxy_url
+        else:
+            destination_url = transfer.transfer_url
+        if direction == otypes.ImageTransferDirection.DOWNLOAD:
+            client.download(
+                destination_url,
+                module.params['download_image_path'],
+                auth.get('ca_file'),
+                fmt='qcow2' if module.params['format'] == 'cow' else 'raw',
+                incremental=module.params['format'] == 'cow',
```
nirs, line 386: incremental is relevant only to incremental backup, which is not related to the format argument. Please remove this.

plugins/modules/ovirt_disk.py
```diff
@@ -366,27 +372,27 @@ def transfer(connection, module, direction, transfer_func):
...
+                secure=not auth.get('insecure'),
+            )
+        else:
+            client.upload(
+                module.params['upload_image_path'],
+                destination_url,
+                auth.get('ca_file'),
+                secure=not auth.get('insecure'),
+            )
```
nirs, line 395: This looks correct.

mnecas:
I have implemented the `os.path.getsize` to get the size of the image instead of `qemu-img info` do you think there would be problems with it?

nirs:
Yes, os.path.getsize() is correct only for raw images. For qcow2 images you need to run:

   qemu-img info --output json  <path>

and fetch "virtual-size" key.

For the initial_size you need to run:

    qemu-img measure -f <src format> -O <target format> <src path> --output json

And use the "required" value.

All this and more (e.g. detecting ISO files) is in upload_disk.py.

requirements.txt
```diff
@@ -1 +1,3 @@
 ansible>=2.9.0
+python3-ovirt-engine-sdk4>=4.4.0
```
ganto, line 2: That's not a valid module name. I think this should be `ovirt-engine-sdk-python>=4.4.0`

requirements.txt
```diff
@@ -1 +1,3 @@
...
+ovirt-imageio-client>=2.0.5
```
ganto, line 3: There is no such module in pypi. Are you still planning to release that? Otherwise it would be good to use the code from #35 as fallback.

requirements.txt
mnecas: Good point

requirements.txt
mnecas: Do you have it in plan @nirs?

nirs: Currently we don't publish ovirt-imgeio in pypi. We can do this but it will not work because it depends on qemu-img rpm, so I don't think this is useful.

We don't plan to support a fallback when qemu-img is not available since it means loss of major functionality. The proper way to use imageio is to install the rpm.

If the ansible script  wants to provide minimal functionallity on a system without the required packages you can have a fallback code that does dump upload and download. But you are basically on your own writing such code since we provided a package that does the right thing, and we don't have capacity to support people who do not use our packages.

mwperina: So would it be possible to have current code as a fallback? Will it work with 4.4 or do we need to detect engine version and act accordingly?

mnecas:
Update:
With @mwperina we have decided
- to drop official support for PyPI https://github.com/oVirt/ovirt-ansible-collection/pull/47
- and wait for the ovirt-img tool so we don't need to update the ansible module with every change. I'll close this PR and when the tool will be ready I'll create another. https://bugzilla.redhat.com/show_bug.cgi?id=1626262

I have done some testing (on 4.3 and 4.4) and will continue with PR https://github.com/oVirt/ovirt-ansible-collection/pull/35.
The reason for this is that it does not add any other dependencies and we could backport it to ansible 2.9 without a problem.
On that note, I will need to do little bit changes in the PR and split it to two (the maintainers of ansible would not allow adding a new parameter (`use_proxy`) on minor release).

@ganto really great job and thank you for the PR.

PR #45, opened 2020-05-14
hard code nic on_boot to True

### LINKED ISSUE #32: nic_on_boot boolean cannot be false
As per https://github.com/oVirt/ovirt-engine/blob/5b6a06433a5c352b53217e75b3219f84f4d41095/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/VmInitToOpenStackMetadataAdapter.java#L48 it seems that  the cloud-init option to set nic_on_boot to false is not valid.

When I attempt to do so I see:

```"msg": "Fault reason is \"Operation Failed\". Fault detail is \"[Cannot add VM. Start On Boot selection in cloud-init configuration is invalid. Only 'true' is supported.]\". HTTP response code is 400."```

snecklifter:
False is not supported and results in an error so this should be
hard-coded to True.

Fixes #32

### DIFF
--- a/plugins/modules/ovirt_template.py
+++ b/plugins/modules/ovirt_template.py
@@ -239,10 +239,6 @@ options:
             nic_name:
                 description:
                     - Set name to network interface of Virtual Machine.
-            nic_on_boot:
-                description:
-                    - If I(True) network interface will be set to start on boot.
-                type: bool
     cloud_init_nics:
         description:
             - List of dictionaries representing network interfaces to be setup by cloud init.
@@ -265,10 +261,6 @@ options:
             nic_name:
                 description:
                     - Set name to network interface of Virtual Machine.
-            nic_on_boot:
-                description:
-                    - If I(True) network interface will be set to start on boot.
-                type: bool
     ballooning_enabled:
         description:
             - "If I(true), use memory ballooning."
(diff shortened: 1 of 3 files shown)


ovirt-infra:

mnecas:
ci add to whitelist

mwperina:
We can't just remove existing option, it would break backward compatibility. We need to deprecate it and keep it for at least some time

mnecas:
> We can't just remove existing option, it would break backward compatibility. We need to deprecate it and keep it for at least some time

This is removing the only suboption, it will not fail/effect when switching from an older version of the modules. 

LGTM

dominikholler:
What is the behavior of this PR if `nic_on_boot: false`?
Maybe we could have at a warning, that this option is ignored, if it is set to false?

snecklifter:
@dominikholler because setting to `nic_on_boot: false` has never, AFAICT, worked. E.g. https://bugzilla.redhat.com/show_bug.cgi?id=1424786 so this is just removing the option (and therefore the confusion/complexity). A warning would only serve to mislead as its not supported by ovirt at all.

snecklifter:
> We can't just remove existing option, it would break backward compatibility. We need to deprecate it and keep it for at least some time

Yes, this doesn't break anything, just removes the broken option from being exposed to the user.

dominikholler:
What happens with your patch if `nic_on_boot: true` is set? Is this ignored?

mnecas:
Yes, it will not fail because the parameter is the only suboption wich ansible does not enforce.

dominikholler:
I see, thanks.
LGTM

PR #58, opened 2020-06-11
Set inventory plugin insecure if no cafile defined

seansackowitz:
In Tower 3.7, I was running into an issue with a self signed certificate on my RHV cluster. Disabling secure connections when the cafile is not defined fixes this issue, in the same way that it worked in previous versions of Tower with the ovirt4 inventory plugin.

### DIFF
--- a/plugins/inventory/ovirt.py
+++ b/plugins/inventory/ovirt.py
@@ -244,7 +244,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             username=self.get_option('ovirt_username'),
             password=self.get_option('ovirt_password'),
             ca_file=self.get_option('ovirt_cafile'),
-            insecure=self.get_option('ovirt_insecure'),
+            insecure=not self.get_option('ovirt_cafile'),
         )
 
         query_filter = self._get_query_options(self.get_option('ovirt_query_filter', None))
--- a/plugins/inventory/ovirt.py
+++ b/plugins/inventory/ovirt.py
@@ -47,7 +47,6 @@ DOCUMENTATION = '''
       ovirt_insecure:
         description: A boolean flag that indicates if the server TLS certificate and host name should be checked.
         required: False
-        default: False
       ovirt_query_filter:
         required: False
         description: dictionary of filter key-values to query VM's. See U(https://ovirt.github.io/ovirt-engine-sdk/master/services.m.html#ovirtsdk4\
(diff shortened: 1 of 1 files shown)


ovirt-infra:

mnecas:
Hi @seansackowitz,

this is a bit gray area I have talked about it with the creator of this plugin in [1].
Please check it out. I'm opened up to further discussion/any notes.

[1] https://github.com/ansible/ansible/pull/49085/files#r377061353
cc @bverschueren

seansackowitz:
While I do agree with the points brought up in that discussion, I still find that the overall experience as a user is failing to meet expectations. And by that I mean, when using the provided RHV credential type in Tower, which doesn't require the ca to be specified, but doesn't allow you to set OVIRT_INSECURE, you are unable to utilize the provided inventory source for ovirt, as it is requiring the ca file by default, and there is no option to override that requirement.

So, I see two solutions:
1. Edit https://github.com/ansible/awx/blob/devel/awx/main/models/credential/__init__.py#L1042 to enable setting the insecure flag in the credential itself
2. Make the change here to follow convention even within this repository: https://github.com/oVirt/ovirt-ansible-collection/blob/master/plugins/module_utils/ovirt.py#L377

mnecas:
@seansackowitz good point that it breaks backwards compatibility with the older plugin and user cannot edit it. 
To be honest I would probably continue with your suggestion.

plugins/inventory/ovirt.py
```diff
@@ -244,7 +244,7 @@ def parse(self, inventory, loader, path, cache=True):
             username=self.get_option('ovirt_username'),
             password=self.get_option('ovirt_password'),
             ca_file=self.get_option('ovirt_cafile'),
-            insecure=self.get_option('ovirt_insecure'),
+            insecure=not self.get_option('ovirt_cafile'),
```
mnecas, line 247: How about to use insecure when specified and if not default to cafile?
```python
insecure=self.get_option('ovirt_insecure') if self.get_option('ovirt_insecure') is not None else not self.get_option('ovirt_cafile')
```

seansackowitz: I just tested your suggestion, but it appears that ovirt_insecure is being defaulted to false somewhere else.

mnecas: I did a little bit of investigation and the default value is added in the DOCUMENTATION.
I removed the default value and worked as it should.
https://github.com/oVirt/ovirt-ansible-collection/blob/44e9c32b8e47643f95ad142ca6a8dd1a823755f6/plugins/inventory/ovirt.py#L47-L50

bverschueren:
If I recall correctly, default variables are set by specifying it in doc_fragments.

I initially followed the philosophy of the SDK where `insecure` and `ca_file` are decoupled as I thought toggling insecure mode on the presence of a ca_file as a bit awkward. In this scenario however it makes sense.

Please think about updating the documentation too as this change has an impact on the plugins behaviour.

PR #61, opened 2020-06-16
ovirt_disk: fix activate

mnecas:
Issue: was not able to attach the deactivated disk to the VM
This keeps previous logic the same (default is activated) but when specifying the value it will overwrite it.
@mwperina @dangel101

### DIFF
--- a/plugins/modules/ovirt_disk.py
+++ b/plugins/modules/ovirt_disk.py
@@ -724,7 +724,7 @@ def main():
         if state in ('present', 'detached', 'attached'):
             # Always activate disk when its being created
             if vm_service is not None and disk is None:
-                module.params['activate'] = True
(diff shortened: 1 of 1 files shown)


gcecchi:
I confirm that it worked for me, both directly using ansible-playbook command from a CentOS 7.8 controller node with ansible-2.9.9-1.el7.noarch, after patching /usr/lib/python2.7/site-packages/ansible/modules/cloud/ovirt/ovirt_disk.py and also using job template in an AWX environment with ansible 2.8.5 after patching the same way the custom venv file /opt/my-envs/py36/lib/python3.6/site-packages/ansible/modules/cloud/ovirt/ovirt_disk.py. I put a pause of 10 seconds between attach task (with actvate: no) and activation task and I have the time to see in web admin gui the disk first attached with the still red arrow and then becoming green. Thanks for the quick fix

PR #67, opened 2020-07-22
inventory plugin: Check if stat values is not empty

machacekondra:


machacekondra:
Oh, that's mistake, actually current code is just fine. Closing PR.

PR #72, opened 2020-07-29
Split to rhv and ovirt builds

mnecas:
I have updated 
- every file which needs to be a little bit different between upstream and downstream `@NAMESPACE@.@NAME@`.
- `build.sh` to automatically convert the `@NAMESPACE@.@NAME@` to `redhat.rhv` or `ovirt.ovirt`
- automation for this kind of behavior - creates two builds in folders `ovirt-build` and `rhv-build`

From now on when someone will be testing they will need to run the build. 
Example:
1. in project ovirt-ansible-collection run `./build.sh build ovirt ~/.ansible/collections/ansible_collections/ovirt/ovirt`


mnecas:
@dangel101 @mwperina @sandrobonazzola

sandrobonazzola:
Is it really needed to differentiate between rhv and ovirt here? I believe it should be fine on RHV to see things referring to ovirt.

mnecas:
It would work without a problem to have everywhere ovirt.ovirt but the reason why we need to do this is the Automation hub where we will be under namespace redhat and will have name rhv.

mnecas:
ci test please

mnecas:
ci test please

mwperina (approved):
+1

PR #84, opened 2020-08-12
Fix changelogs

mnecas:
Fix changelogs folder from now when doing some PR add info to fragments and when releasing bump galaxy version and `antsibull-changelog release` to generate logs.
#74


mnecas:
tests fail on missing dependency will solve it in #72

PR #97, opened 2020-09-01
Add role: manageiq

mnecas:
Migrating the manageiq role from commit: https://github.com/oVirt/ovirt-ansible-manageiq/commit/f91b9061efd3d61cdc8a6b117cabd676e9dbab43


mnecas:
ci test please

mwperina (approved):
It looks good, but in short future we should bump the defaults to Ivanchuk, which is the latest stable release

PR #119, opened 2020-09-15
roles: hosted_engine_setup: Use the correct FQCN

arachmani:


mnecas (approved):
LGTM

mwperina (commented):
Hmm, shouldn't we use `@COLLECTION_NAMESPACE@.@NAME@` so it would be properly set for both upstream and downstream during the build?

mnecas:
> Hmm, shouldn't we use `@COLLECTION_NAMESPACE@.@NAME@` so it would be properly set for both upstream and downstream during the build?

Yes, that's correct.

PR #120, opened 2020-09-15
ovirt_disk: fix upload when direct upload fails

mnecas:
There were multiple possible exceptions when trying to connect to the engine for direct upload to the host.
This fixes the issue and solves any other unexpected errors.
https://bugzilla.redhat.com/show_bug.cgi?id=1879165


mwperina (approved):
+1

PR #127, opened 2020-09-22
infra, engine_setup: remove collections

mnecas:
infra, engine_setup: remove collections
This fixes the usage of modules from inside the collection


mwperina (approved):
+1

PR #134, opened 2020-10-05
dr: add Disaster Recovery project to collections

barpavel:
Move the the Disaster Recovery project from the
https://github.com/oVirt/ovirt-ansible-disaster-recovery
to the "ovirt-ansible-collection" with the relevant modifications.

Signed-off-by: Pavel Bar <pbar@redhat.com>


roles/disaster_recovery/README.md
```diff
@@ -0,0 +1,93 @@
+oVirt Disaster Recovery
+=========
+
+The `oVirt.disaster-recovery` role responsible to manage the disaster recovery scenarios in oVirt.
```
mnecas, line 4: Rename the `oVirt.disaster-recovery` to  `disaster_recovery`

roles/disaster_recovery/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+Requirements
+------------
+
+ * Ansible version 2.9 or higher
+ * Python SDK version 4.4 or higher
```
mnecas, line 10: There are global requirements in the main README.md. You can remove these lines.

roles/disaster_recovery/README.md
```diff
@@ -0,0 +1,93 @@
...
+| dr_clean_orphaned_disks        | True                  | Specify whether to remove lun disks from the setup as part of engine setup.       |
+| dr_running_vms		 | /tmp/ovirt_dr_running_vm_list	 | Specify the file path which is used to contain the data of the running VMs in the secondary setup before the failback process run on the primary setup after the secondary site cleanup was finished. Note that the /tmp folder is being used as default so the file will not be available after system reboot.
+
+
+
+Dependencies
+------------
+
+No.
+
```
mnecas, line 35: There are global dependencies in the main README.md. You can remove these lines.

roles/disaster_recovery/README.md
```diff
@@ -0,0 +1,93 @@
...
+```yaml
+---
+- name: Setup oVirt environment
+  hosts: localhost
+  connection: local
+  vars_files:
+     - ovirt_passwords.yml
+     - disaster_recovery_vars.yml
+  roles:
+     - {role: oVirt.disaster-recovery}
```
mnecas, line 48: Change
```
     - {role: oVirt.disaster-recovery}
```
to 
```yaml
  roles:
    - disaster_recovery
  collections:
    - ovirt.ovirt
```

roles/disaster_recovery/README.md
```diff
@@ -0,0 +1,93 @@
...
+For fail-back operation:
+```console
+$ ./ovirt-dr failback
+```
+
+License
+-------
+
+Apache License 2.0
+
```
mnecas, line 93: Remove License

roles/disaster_recovery/README.md
```diff
@@ -0,0 +1,93 @@
...
+Each of those actions are using a configuration file whose default location is `oVirt.disaster-recovery/files/dr.conf`<br/>
+The configuration file's location can be changed using `--conf-file` flag in the `ovirt-dr` script.<br/>
+Log file and log level can be configured as well through the `ovirt-dr` script using the flags `--log-file` and `--log-level`
+
+
+Example Script
+--------------
+For mapping file generation:
+```console
+$ ./ovirt-dr generate --log-file=ovirt-dr.log --log-level=DEBUG
```
mnecas, line 74: Probably change the path of the ovirt-dr

roles/disaster_recovery/callback_plugins/stdout.py
```diff
@@ -0,0 +1,57 @@
+from datetime import datetime
```
mnecas, line 1: I think this should be moved to the plugins of the collection. 
Create `callback` folder in https://github.com/oVirt/ovirt-ansible-collection/tree/master/plugins and move it there.

roles/disaster_recovery/examples/generator.log
```diff
@@ -0,0 +1,2152 @@
+DEBUG:root:* Trying 192.168.1.248:8080...
```
mnecas, line 1: Not sure if we should have the log in the collection.
Is it a necessary example?

roles/disaster_recovery/tasks/generate_mapping.yml
```diff
@@ -0,0 +1,6 @@
+- block:
+    - name: Generate mapping var file
+      command: python3 {{ role_path }}/files/generate_mapping.py -a "{{ site }}" -u "{{ username }}" -p "{{ password }}" -c "{{ ca }}" -f "{{ var_file }}"
```
mnecas, line 3: Watch out if the `role_path` is correct

mnecas (changes requested):
Hi,
thanks for the PR wrote some notes.
If you want to test the collection please this command:
`build.sh build ovirt ~/.ansible/collections/ansible_collections/ovirt/ovirt`
it will generate collection and you will be able to use it as in examples.

Please also add the fragment you can read about it in: 
https://github.com/oVirt/ovirt-ansible-collection/tree/master/changelogs

roles/disaster_recovery/README.md
barpavel: Thanks for this and other comments.
I wanted to address the renaming in the follow-up patches.
Deliberately wanted this patch to be a "clean" copy of the original :)

mwperina: Leaving ```-``` in role name would break collections, so I'd prefer to have single commit to import current status and then additional commit to fix issues. But all commits need to be within a single PR

barpavel: Yes, I rephrase myself :)
I meant follow-up **commits** (not patches). Commits inside **this patch**.
All I wanted was to have the 1st commit for just copying DR project logic (something that I already see didn't work for 100% since the CI failed). And the next ones to contain all the modifications needed to support collections.
For clear separation of all the changes done due to collections.

barpavel: > oVirt.disaster-recovery

Done.

roles/disaster_recovery/README.md
barpavel: Done.

roles/disaster_recovery/README.md
barpavel: Done, removed.

roles/disaster_recovery/README.md
barpavel: Done, removed.

roles/disaster_recovery/tasks/generate_mapping.yml
barpavel: I will still need to test the full flow, so it will simply not work if the `role_path` is not correct. The more interesting question is whether it's correct to execute `python3` directly.

roles/disaster_recovery/README.md
barpavel: Why should I change it?
This is an example. In this example the log file `ovirt-dr.log` will be created in the current folder. This parameter may be omitted at all, in that case the created log file will be located under `/tmp/`.
Or maybe I didn't understand what you suggested :)

barpavel:
> Please also add the fragment you can read about it in:
> https://github.com/oVirt/ovirt-ansible-collection/tree/master/changelogs

Fragment was added.

mwperina:
It looks good to me, isn't it worth to remove WIP?

roles/disaster_recovery/examples/dr_play.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Setup oVirt environment
+  hosts: localhost
+  connection: local
+  roles:
+     - disaster_recovery
```
mnecas, line 6: missing collection usage( `collections: ovirt.ovirt` )

roles/disaster_recovery/examples/dr_ovirt_setup.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Setup oVirt environment
+  hosts: localhost
+  connection: local
+  vars_files:
+     - ovirt_passwords.yml
+     - disaster_recovery_vars.yml
+  roles:
+     - disaster_recovery
```
mnecas, line 9: missing collection usage( collections: ovirt.ovirt )

mnecas (changes requested):
Just a few small nitpicks

roles/disaster_recovery/callback_plugins/stdout.py
mnecas: probably misunderstood my comment.
I was referring to the callback plugin, not the import.
I think this you should move the whole file to the `plugins/callback`.

roles/disaster_recovery/README.md
mnecas: Previously the `ovirt-dr` was on top of the project. Now the script is under `roles/disaster_recovery/` so might be better to change the path to something like`./roles/disaster_recovery/ovirt-dr` just suggestion. I'm open for any dissgusion

roles/disaster_recovery/tasks/generate_mapping.yml
mnecas: yeah, I was just pointing out that it can fail here because it does not know it's under `roles/disaster_recovery`
(maybe I'm just missing something)

roles/disaster_recovery/README.md
barpavel: > Remove License

BTW: I see that "hosted_engine_setup" still has the licence line at the end of the readme file. So should this line be removed?

roles/disaster_recovery/README.md
barpavel: Done.
Handled in the commit that handles all the `README.md` changes.

roles/disaster_recovery/callback_plugins/stdout.py
barpavel: Done.
No misunderstanding, just didn't handle till now :)
Anyway, committed a separate commit named "moving "stdout.py" callbacks file".

barpavel:
> It looks good to me, isn't it worth to remove WIP?

Moved to "code review ready" state :)
Now handling the remaining comments from Martin Necas.

roles/disaster_recovery/examples/dr_play.yml
barpavel: Done.
Commit "updating the role name" was updated with the fix.

roles/disaster_recovery/examples/dr_ovirt_setup.yml
barpavel: Done.
Commit "updating the role name" was updated with the fix.

roles/disaster_recovery/tasks/generate_mapping.yml
barpavel: I'm starting testing soon. But the question is should I test it on Python2 version? Because the current implementation calls the `python3` command...

mwperina: It depends where the role is going to be executed. If you want only EL8 support, then test Python 3 only. If you expect, that role will be executed even on older 4.3/4.2 hosts, then you need to support both Python 2 and Python 3

mnecas: I think there should be another PR for this and not do everything when migrating.

mwperina (approved):
Looks good to me, let's merge it so we can start testing from collection

mnecas (approved):
LGTM

PR #138, opened 2020-10-07
repositories: fix gpg key

mnecas:
The `disable_gpg_check` was added because of the master rpm are not signed by the gpg key
@mwperina


mwperina (approved):
+1

PR #141, opened 2020-10-09
[WIP] Add builds for s390x architecture

mwperina:


mwperina:
it doesn't make sense to introduce specific build jobs for s390x arch, because ovirt-ansible-collection is noarch package, so it's installable on all platforms. So abandoning this PR.

PR #148, opened 2020-10-15
fix the build path to copy all files into correct path

kobihk:
maybe it will be better to use the BUILD_PATH as a base path (where all collections should located).
the same as:
    ANSIBLE_COLLECTIONS_PATH

and add the following line to build script: 
    BUILD_PATH=$BUILD_PATH/ansible_collections/ovirt/ovirt/ 
instead of everyone add it in his build command.
the build command will look like:
    ./build.sh build ovirt "$ANSIBLE_COLLECTIONS_PATH"

### DIFF
--- a/automation/build-artifacts.sh
+++ b/automation/build-artifacts.sh
@@ -13,7 +13,7 @@ mkdir -p ../ovirt-build ../rhv-build
 ./build.sh build ovirt ../ovirt-build
 ./build.sh build rhv ../rhv-build
 
-cd ../ovirt-build
+cd $ROOT_PATH/ovirt-build/ansible_collections/ovirt/ovirt/
 # create the src.rpm
 rpmbuild \
     -D "_srcrpmdir $PWD/output" \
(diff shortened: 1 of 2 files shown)


ovirt-infra:

mnecas:
ci add to whitelist

mwperina (approved):
+1

automation/build-artifacts.sh
```diff
@@ -3,17 +3,18 @@
 ROOT_PATH=$PWD
 
 # remove any previous artifacts
-rm -rf ../ovirt-build ../rhv-build
+rm -rf ../ansible_collections
```
kobihk, line 6: if there are other ansible collections do we want to delete them too?

mnecas: this is just in CI automation which tests the collection
and when it reruns it has the previous build in it

kobihk: what I meant when I have a few other collections that already installed 
I think if we will remove the folder ansible_collections
we can remove the ones that we already installed but if we will be more specific like:
 rm -rf ../ansible_collections/ovirt  ../ansible_collections/redhat
we can be saved for both...wdyt?

mnecas: Yes, you are right but this change does not affect the build script. When you run the build script it will not remove the previous collections.
The automation folder is only for tests of the collection inside Jenkins.

kobihk: ok, thanks!!

PR #143, opened 2020-10-11
engine_setup: Don't update ansible package

arachmani:
Don't update ansible package as we have version lock
on this package

Bug-Url: https://bugzilla.redhat.com/1887142
Signed-off-by: Asaf Rachmani <arachman@redhat.com>

PR #149, opened 2020-10-19
engine_setup: yum.conf set 'best' to false

mnecas:
Continuation of https://github.com/oVirt/ovirt-ansible-collection/pull/143
@mwperina


mwperina (approved):
+1

PR #151, opened 2020-10-20
roles: hosted_engine_setup: Improve engine logs fetching

didib:
This is an addition to [1], missing there, updating also
partial_execution.yml . That said:

- This does not (usually?) reproduce on CI, perhaps simply because we
run there with low memory (thus forcing the kernel to sync often)

- I do not know well which parts of this code run on the host, which on
the engine, and what and when adds the engine to the inventory, so not
sure it would always work.

Copying the commit message from [1]:

Sometimes, on failed deploy attempts, engine-logs* directories are
created but left empty. I am guessing that this might be because the
files there are still cached in memory, and we do not fetch from the
engine vm itself (using e.g. ssh) but from the host, directly accessing
the vm disk image. So if, before that, we run 'sync' on the engine vm,
we have better chances to successfully get the logs from the image.

[1] https://github.com/oVirt/ovirt-ansible-hosted-engine-setup/pull/325

Bug-Url: https://bugzilla.redhat.com/1844965

### DIFF
--- a/roles/hosted_engine_setup/tasks/add_engine_as_ansible_host.yml
+++ b/roles/hosted_engine_setup/tasks/add_engine_as_ansible_host.yml
@@ -1,0 +1,24 @@
+---
+- name: Add the engine VM as an ansible host
+  block:
+  - name: Fetch the value of HOST_KEY_CHECKING
+    set_fact: host_key_checking="{{ lookup('config', 'HOST_KEY_CHECKING') }}"
+  - debug: var=host_key_checking
+  - name: Get the username running the deploy
+    become: false
+    command: whoami
+    register: username_on_host
+  - name: Register the engine FQDN as a host
+    add_host:
+      name: "{{ he_fqdn }}"
+      groups: engine
+      ansible_connection: smart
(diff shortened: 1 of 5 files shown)


mnecas:
@arachmani can you please review?

arachmani (approved):
Looks good to me

didib:
> Looks good to me

Good enough so that I should remove from the commit message the second bullet :-) ?

I don't mind keeping it, though. Yesterday I spent quite some time trying to understand the current structure of this project, how it works, etc., including reading about the change from include_roles to import_roles, etc., and decided to give up for now, and just push this patch. Still working on verification.

didib:
This is broken:

[ INFO  ] TASK [ovirt.ovirt.hosted_engine_setup : Sync on engine machine]
[ ERROR ] fatal: [localhost]: FAILED! => {"msg": "'dict object' has no attribute 'engine'"}

I guess I'll have to learn a bit more...

sandrobonazzola (approved):
to my limited knowledge of anisble, looks good to me

arachmani (approved):

PR #166, opened 2020-10-28
infra: fix Python3 support - iteritems

mnecas:
Backport of https://github.com/oVirt/ovirt-ansible-infra/pull/81

### DIFF
--- a/roles/infra/roles/aaa_jdbc/tasks/main.yml
+++ b/roles/infra/roles/aaa_jdbc/tasks/main.yml
@@ -34,7 +34,7 @@
     - users
 
 - name: Update users according to attributes
-  command: "{{ aaa_jdbc_prefix }}/ovirt-aaa-jdbc-tool user edit {{ item.name }} {% for attr, value in item['attributes'].iteritems() %} --attribute={{ attr }}='{{ value }}' {% endfor %}"
+  command: "{{ aaa_jdbc_prefix }}/ovirt-aaa-jdbc-tool user edit {{ item.name }} {% for attr, value in item['attributes'].items() %} --attribute={{ attr }}='{{ value }}' {% endfor %}"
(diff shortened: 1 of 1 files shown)


mnecas:
ci test please

roles/infra/roles/aaa_jdbc/tasks/main.yml
```diff
@@ -34,7 +34,7 @@
     - users
 
 - name: Update users according to attributes
-  command: "{{ aaa_jdbc_prefix }}/ovirt-aaa-jdbc-tool user edit {{ item.name }} {% for attr, value in item['attributes'].iteritems() %} --attribute={{ attr }}='{{ value }}' {% endfor %}"
+  command: "{{ aaa_jdbc_prefix }}/ovirt-aaa-jdbc-tool user edit {{ item.name }} {% for attr, value in item['attributes'].items() %} --attribute={{ attr }}='{{ value }}' {% endfor %}"
```
mwperina, line 37: According to https://www.python.org/dev/peps/pep-0469/ shouldn't we rather use `iter(item['attributes'].items())` ?
Marcin, what do you think?

@tinez

tinez: It doesn't make that much of a difference here. First of all, we're within an iteration context and not assigning to a variable. `.items()` gives us a list in py2 and py3 gives a view. Even if we use the extra `iter()` layer of wrapping, we've already created the whole list object in memory, so we're missing the point here.

Unless we expect the `item['attributes']` dictionary to be really big LGTM.

mwperina: OK, thanks

mwperina (approved):
+1

PR #168, opened 2020-11-02
Automation: rename ovirt.ovirt to redhat.rhv in all readme.md

mnecas:
This fixes the documentation for AH builds before there were READMEs with upstream examples.
And this also keeps GitHub readmes untouched.


mnecas:
Continuing in #169

Issue #172, opened 2020-11-03
Inventory plugin broken in python2


kdelee:
##### COMPONENT NAME
ovirt inventory plugin

##### STEPS TO REPRODUCE

Run inventory plugin in python2

##### EXPECTED RESULTS
Works, this is still supported and required for rhel7 Ansible Tower and Ansible Tower on openshift

##### ACTUAL RESULTS
```
ansible-inventory 2.9.14
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/var/lib/awx/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /usr/bin/ansible-inventory
  python version = 2.7.5 (default, Mar 20 2020, 17:08:22) [GCC 4.8.5 20150623 (Red Hat 4.8.5-39)]
Using /etc/ansible/ansible.cfg as config file
[WARNING]:  * Failed to parse /tmp/awx_675_2eostks8/ovirt.yml with auto plugin:
'datetime.datetime' object has no attribute 'timestamp'
  File "/usr/lib/python2.7/site-packages/ansible/inventory/manager.py", line 280, in parse_source
    plugin.parse(self._inventory, self._loader, source, cache=cache)
  File "/usr/lib/python2.7/site-packages/ansible/plugins/inventory/auto.py", line 58, in parse
    plugin.parse(inventory, loader, path, cache=cache)
  File "/var/lib/awx/vendor/awx_ansible_collections/ansible_collections/ovirt/ovirt/plugins/inventory/ovirt.py", line 264, in parse
    source_data = self._query(query_filter=query_filter)
  File "/var/lib/awx/vendor/awx_ansible_collections/ansible_collections/ovirt/ovirt/plugins/inventory/ovirt.py", line 149, in _query
    return [self._get_dict_of_struct(host) for host in self._get_hosts(query_filter=query_filter)]
  File "/var/lib/awx/vendor/awx_ansible_collections/ansible_collections/ovirt/ovirt/plugins/inventory/ovirt.py", line 129, in _get_dict_of_struct
    'creation_time_timestamp': vm.creation_time.timestamp(),
```

### Notes

Just sanity checking that this is a python2 vs python3 problem:
```
$ python2
Python 2.7.18 (default, Jul 20 2020, 00:00:00) 
[...]

PR #173, opened 2020-11-04
Inventory plugin: fix timestamp for python2

mnecas:
Fixes: https://github.com/oVirt/ovirt-ansible-collection/issues/172

Example of output: 
```
                "creation_time_timestamp": "1598263100.532000",
```
Test on python2:
```python
>>> datetime.datetime.now().strftime("%s.%f")
'1604486628.612790'
```

### DIFF
--- a/plugins/inventory/ovirt.py
+++ b/plugins/inventory/ovirt.py
@@ -126,7 +126,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             'os_type': vm.os.type,
             'template': self.connection.follow_link(vm.template).name,
             'creation_time': str(vm.creation_time),
-            'creation_time_timestamp': vm.creation_time.strftime("%s.%f"),
+            'creation_time_timestamp': float(vm.creation_time.strftime("%s.%f")),
             'tags': [tag.name for tag in tags],
             'affinity_labels': [label.name for label in labels],
             'affinity_groups': [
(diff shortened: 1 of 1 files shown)


mwperina:
@tinez

plugins/inventory/ovirt.py
```diff
@@ -126,7 +126,7 @@ def _get_dict_of_struct(self, vm):
             'os_type': vm.os.type,
             'template': self.connection.follow_link(vm.template).name,
             'creation_time': str(vm.creation_time),
-            'creation_time_timestamp': vm.creation_time.timestamp(),
+            'creation_time_timestamp': vm.creation_time.strftime("%s.%f"),
```
tinez, line 129: This changes the type from `float` to `str` - I don't have the bigger picture, so not sure if that's ok. Nit - I can also see more single quoted strings throughout the module (although the quotes are not 100% consistent, so up to you).

kdelee: I suppose you could just wrap it in `float(vm.creation_time.strftime("%s.%f"))`  that way python3 does not change behavior and python2 will have same behavior

tinez: > I suppose you could just wrap it in `float`...

Yes, this will give you the same type for both py2 and py3, but looks really wrong - we're going from numeric type, converting it to string and then back to numeric one. Floating point numbers' representation in memory is power-of-2-based and not power-of-10-based, so parsing from strings can lead to unpredictable results.

AFAICS, this code was introduced [here](https://github.com/oVirt/ovirt-ansible-collection/commit/f86f49730f89d7606f417bd440da85d74c7a2c90), meaning that this is not a regression on py2 - it simply never worked.

Some important questions that would need an answer:
- how is that timestamp used? can we introduce breakage of type and/or precision here?
- do we absolutely, positively need microsecond precision?

One solution I see as reasonable is to use `time.mktime(vm.creation_time.timetuple())`, but the `time_struct` that is created in between doesn't have the microseconds part, so we end up with zero-rounded floats, i.e. `1598263100.0`.

mwperina (approved):
+1

PR #177, opened 2020-11-11
Add "Provides:" to the packages we replace

didib:
We added "Obsoletes:" recently but it seems to not be enough.


mwperina (changes requested):
It's required to manually modify existing playbooks in order to use roles in the collection. That's why I don't think it's good to add provides. Just adding obsoletes make the required effort more visible.
Please take a look at discussion in original PR: https://github.com/oVirt/ovirt-ansible-collection/pull/115

didib:
Fine for me to close this PR - I well understand the implications. But we do have to find a solution to the upgrade to 4.4.3.

mwperina (approved):
+1 to minimize issues during upgrade

PR #180, opened 2020-11-16
Missing restore task file and vars file

kobihk:
when running restore flow we failed with the following error:
TASK [ovirt.ovirt.engine_setup : Restore engine from file] *********************
fatal: [example.com]: FAILED! => {"reason": "Could not find or access
'/path/to/file/restore_engine_from_file.yml' on the Ansible Controller."}

### DIFF
--- a/roles/engine_setup/tasks/restore_engine_from_file.yml
+++ b/roles/engine_setup/tasks/restore_engine_from_file.yml
@@ -1,0 +1,19 @@
+---
+- name: Run engine cleanup command
+  command: "engine-cleanup"
+  when: ovirt_engine_setup_restore_engine_cleanup
+
+- name: Add scopes to restore engine command
+  set_fact:
+    restore_cmd: "{{ restore_cmd }} --scope={{ item }}"
+  with_items: "{{ ovirt_engine_setup_restore_scopes | default([]) }}"
+
+- name: Add restore file and restore options
+  set_fact:
+    restore_cmd: "{{ restore_cmd }} --{{ item.key }}{% if item.value %}={{ item.value }}{% endif %}"
+  with_dict:
+    - file: "{{ ovirt_engine_setup_restore_file }}"
+    - "{{ ovirt_engine_setup_restore_options | default({}) }}"
+
+- name: Run restore engine from backup file
+  command: "{{ restore_cmd }}"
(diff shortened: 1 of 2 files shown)


mwperina:
@didib

mwperina (commented):
Looks good to me, but I would Didi to take a look

roles/engine_setup/tasks/restore_engine_from_file.yml
```diff
@@ -0,0 +1,19 @@
+---
+- name: Run engine cleanup command
+  command: "engine-cleanup"
```
didib, line 3: Without an answerfile? It would get stuck asking questions. Was this ever tested?

Perhaps also add 'ignore_errors: true'.

didib (commented):
We do not use the restore in hosted-engine (we have there our own code for that. Perhaps worth considering merging, not that important right now), so I do not know current code well. It looks ok (other than engine-cleanup). I assume it was a simple omission when copying it from ovirt-ansible-engine-setup. Did you check the current diff between them? There are a few other things not copied - perhaps none are important, but better make sure - things around tests, meta, requirements, etc.

roles/engine_setup/tasks/restore_engine_from_file.yml
kobihk: yes you right!!
we have an open issue about it in:
https://github.com/oVirt/ovirt-ansible-engine-setup/issues/93

kobihk:
The content of this PR merged in:
https://github.com/oVirt/ovirt-ansible-engine-setup/pull/87/files
but I assume by mistake ommited in:
https://github.com/oVirt/ovirt-ansible-collection/pull/113

didib:
Fine for me to merge current, if you also keep tracking open issues in the old project (or copy them here).

mnecas (approved):
LGTM, sorry for missing this when migrating

roles/engine_setup/tasks/restore_engine_from_file.yml
mnecas: I'll move the issue to the collection so we have everything here. (Please let me know if I missed some important issue which should be moved here)

mnecas: https://github.com/oVirt/ovirt-ansible-collection/issues/182

Issue #193, opened 2020-12-10
ovirt inventory plugin doesn't properly filter based on os_type


lunarnexus:
##### SUMMARY
When using the ovirt ansible inventory plugin, ovirt_query_filter: search: "os_type=" doesn't filter properly.

##### COMPONENT NAME

ovirt ansible inventory plugin

##### STEPS TO REPRODUCE

Create Inventory File as ovirt.yml:
```

plugin: ovirt.ovirt.ovirt
ovirt_url: https://hosted-enginel/ovirt-engine/api
ovirt_cafile: ca.pem
ovirt_username: admin@internal
ovirt_password: password
```

View results with:
`ansible-inventory -i ovirt.yml --list`

This command returns results which show vm info such as this excerpt:

```
      },
      "testserver.local": {
        "affinity_groups": [],
        "affinity_labels": [],
        "ansible_host": "10.1.1.10",
        "cluster": "Default",
        "creation_time": "2020-11-19 12:37:50.691000-05:00",
        "creation_time_timestamp": 1605807470.691,
        "description": "Test Server",
        "devices": {
          "eth0": [
            "10.1.1.10",
          ]
        },
        "fqdn": "testserver.local",
        "host": "ovirthost-1",
        "id": "abcdefg",
        "name": "testserver.local",
        "os_type": "rhel_8x64",
        "status": "up",
        "tags": [
        ],
        "template": "Blank"
      },
```

The os_type attribute is populated.  But when using another inventory file with the os_type filter, no results are returned:
```

plugin: ovirt.ovirt.ovirt
ovirt_url: https://hosted-enginel/ovirt-engine/api
ovirt_cafile: ca.pem
ovirt_username: admin@internal
ovirt_password: password
ovirt_query_filter:
  all_content: true
  search: 'os_type=rhel_8x64'
  case_sensitive: no

```
```
# ansible-inventory -i ovirt.yml --list
{
  "_meta": {
    "hostvars": {}
  },
  "all": {
    "children": [
      "ungrouped"
    ]
  }
}
```

##### EXPECTED RESULTS

```
ovirt_query_filter:
  all_content: true
  search: 'os_type=rhel_8x64'
```

This should return all servers which match this os_type.

mnecas:
Hi,
please change in the search `os_type` to `os`. 
For some reason, the search does not have the `os_type` will check out why.

PR #194, opened 2020-12-11
correct name of os_type

mnecas:
Fixes https://github.com/oVirt/ovirt-ansible-collection/issues/193

The return of the inventory has os_type inside even if we don't use the variable in search.
This breaks bp not sure if we should implement it.


mwperina (approved):
+1

PR #199, opened 2021-01-04
ovirt_host_network: fix bond modes

mnecas:
When rerunning the same playbook there was an issue of redeployment of the same network.
It was caused by a missing config in getting the bond options so the check update function was not returning correct value.

### DIFF
--- a/plugins/modules/ovirt_host_network.py
+++ b/plugins/modules/ovirt_host_network.py
@@ -242,15 +242,6 @@ from ansible_collections.@NAMESPACE@.@NAME@.plugins.module_utils.ovirt import (
 
 
 def get_bond_options(mode, usr_opts):
-    MIIMON_100 = dict(miimon='100')
-    DEFAULT_MODE_OPTS = {
-        '0': MIIMON_100,
-        '1': MIIMON_100,
-        '2': MIIMON_100,
-        '3': MIIMON_100,
-        '4': dict(xmit_hash_policy='2')
-    }
-
     options = []
     if mode is None:
         return options
@@ -284,8 +275,12 @@ def get_bond_options(mode, usr_opts):
             value=str(mode_number)
         )
     )
+    opts_dict = {}
+    if mode == 4:
+        opts_dict = dict(xmit_hash_policy='2')
+    elif mode < 4:
+        opts_dict = dict(miimon='100')
 
-    opts_dict = DEFAULT_MODE_OPTS.get(str(mode), {})
     if usr_opts is not None:
         opts_dict.update(**usr_opts)
 
--- a/plugins/modules/ovirt_host_network.py
+++ b/plugins/modules/ovirt_host_network.py
@@ -244,10 +244,11 @@ from ansible_collections.@NAMESPACE@.@NAME@.plugins.module_utils.ovirt import (
 def get_bond_options(mode, usr_opts):
     MIIMON_100 = dict(miimon='100')
     DEFAULT_MODE_OPTS = {
+        '0': MIIMON_100,
         '1': MIIMON_100,
         '2': MIIMON_100,
         '3': MIIMON_100,
-        '4': dict(xmit_hash_policy='2', **MIIMON_100)
+        '4': dict(xmit_hash_policy='2')
     }
 
     options = []
@@ -264,6 +265,8 @@ def get_bond_options(mode, usr_opts):
             'Load balance (balance-xor)',
             None,
             'Dynamic link aggregation (802.3ad)',
+            'Adaptive transmit load balancing (balance-tlb)',
+            None,
         ]
         if (not 0 < mode_number <= len(modes)):
             return None


plugins/modules/ovirt_host_network.py
```diff
@@ -264,6 +265,8 @@ def get_type_name(mode_number):
             'Load balance (balance-xor)',
             None,
             'Dynamic link aggregation (802.3ad)',
+            'Adaptive transmit load balancing (balance-tlb)',
+            None,
```
mwperina, line 269: Hmm, shouldn't we align with official doc here?

https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.4/html/administration_guide/sect-network_bonding#Bonding_Modes

plugins/modules/ovirt_host_network.py
```diff
@@ -244,10 +244,11 @@
 def get_bond_options(mode, usr_opts):
     MIIMON_100 = dict(miimon='100')
     DEFAULT_MODE_OPTS = {
+        '0': MIIMON_100,
         '1': MIIMON_100,
         '2': MIIMON_100,
         '3': MIIMON_100,
-        '4': dict(xmit_hash_policy='2', **MIIMON_100)
+        '4': dict(xmit_hash_policy='2')
```
mwperina, line 251: Shouldn't we align with the real bond mode numbers?

https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.4/html/administration_guide/sect-network_bonding#Bonding_Modes

Is there any official doc around bond mode default options in oVirt/RHV?

mwperina (commented):
Dominik/Ales could you please take a look?

@dominikholler @almusil

plugins/modules/ovirt_host_network.py
```diff
@@ -264,6 +265,8 @@ def get_type_name(mode_number):
             'Load balance (balance-xor)',
             None,
```
almusil, line 266: Just currious why is the Broadcast name None?

plugins/modules/ovirt_host_network.py
dominikholler: @mnecas can you please help me to understand the purpose of these strings here?

dominikholler (commented):
This code change might be related to https://bugzilla.redhat.com/show_bug.cgi?id=1890071

plugins/modules/ovirt_host_network.py
mwperina: > > Shouldn't we align with the real bond mode numbers?
> 
> @mwperina , can you please explain?

Here are only 5 modes out of 7 existing, so shouldn't we have a full list here?

> 
> > Is there any official doc around bond mode default options in oVirt/RHV?
> 
> No, on REST-API there might be no magic done by RHV at all, and on the host the local defaults apply, but are not reported by RHV. On webadmin, the logic from
> https://github.com/oVirt/ovirt-engine/blob/50129b39cd9315488e3da37446aee7aa65c0d71e/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/network/BondMode.java#L45
> 
> is applied.

OK, so does that mean that we don't need to append `miimon='100'` inside ansible module, because engine performs that automatically? I think all defaults should be handled on engine side and clients likes SDKs and Ansible modules shouldn't have anything copied inside (like here)

plugins/modules/ovirt_host_network.py
mnecas: This just checks what is in API with what user added to ansible. 
Users can choose the mode as a number we add the strings and dictionaries to match with the API response.

plugins/modules/ovirt_host_network.py
mnecas: I did not need to add the other modes because they don't have attached `miimon='100'`.

plugins/modules/ovirt_host_network.py
mnecas: API does not have any name in it, also for the 6. mode.

mwperina: Hmm, looks like we need to fix HostNicMapper and align it with BondMode enum:

https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/restapi/types/src/main/java/org/ovirt/engine/api/restapi/types/HostNicMapper.java#L34
https://github.com/oVirt/ovirt-engine/blob/50129b39cd9315488e3da37446aee7aa65c0d71e/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/network/BondMode.java#L45

dominikholler: Isn't the `type` attribute of the bond option only informative, and the mode is defined by the `value` attribute? Wouldn't it better to use only the `value` and ignore the `type` in this module?

plugins/modules/ovirt_host_network.py
dominikholler: https://github.com/oVirt/ovirt-ansible-collection/pull/199/files#r551842364

plugins/modules/ovirt_host_network.py
dominikholler: The MIIMON_100 is added by Engine UI, so either the ansible module could add it, too to be consistent with the Engine UI, or not to add it, to be consistent with the REST API.

plugins/modules/ovirt_host_network.py
mnecas: I could try to ignore all options with the name `miimon` not sure if it would work. Just noticed it is also used when creating the network.

dominikholler: I do not get the problem, which this code changes tries to solve.

- One possible desired behavior would be to handle the miimon in ansible like any option. This would mean that the user would have to specify the option if he wants it in all modes except 4, and the user should not specify miimon in mode 4, because it will be not reported by oVirt. This would put the work to the user, and would reflect what the user would have to speifcy manually on the host. A problem might be the backward compatibility of the playbooks.
- Another possible desired behavior would be to not report a difference if the miimon is expected to be set, but it is not reported by Engine in mode 4. This would be wrong, if the kernel of the host would remove miimon from defaults, but I guess this will not happen in RHEL8.

I would prefer the first one without the second one, but of course they could be combined.

mnecas: This patch just updates the module so it matches the API. I don't know if the API is correct, that's probably up to you, but I use it as my guidelines for the ansible. 

We don't need to specify the `miimon` from a user. We just get the number of the mode and inside the module, we add to it the required options for the request. This way it does not break backward compatibility.

dominikholler: > This patch just updates the module so it matches the API. I don't know if the API is correct, that's probably up to you, but I use it as my guidelines for the ansible.
> 

oVirt -engine REST API does not set the miimon implicitly.

> We don't need to specify the `miimon` from a user. We just get the number of the mode and inside the module, we add to it the required options for the request. This way it does not break backward compatibility.

It might be convinet for the user

plugins/modules/ovirt_host_network.py
```diff
@@ -281,8 +275,12 @@ def get_type_name(mode_number):
             value=str(mode_number)
         )
     )
+    opts_dict = {}
+    if mode == 4:
+        opts_dict = dict(xmit_hash_policy='2')
+    elif mode < 4:
```
dominikholler, line 281: Why `elif mode < 4:` or why not `else:`?

dominikholler (commented):
Thanks, just one question.

plugins/modules/ovirt_host_network.py
mnecas: I found out that in API the 5. and 6. mode does not have miimon.

mwperina:
ci build

Issue #204, opened 2021-01-22
Provide module to fetch system options from oVirt Engine RESTAPI


mwperina:
We have oVirt Engine RESTAPI endpoint to fetch system options:

http://ovirt.github.io/ovirt-engine-api-model/4.4/#services/system_options

We need to implement module ovirt_system_options_info to be able to fetch those system options. We should be able to support following flows:

1. Provide option name and version to fetch specific value:
```
ovirt_system_options:
    name: "IsPortIsolationSupported"
    version: "4.4"
```
That should return only value for specific version, for example:
```
  'IsPortIsolationSupported': {
    '4.4': 'false'
  }
```
2. Provide only option name to fetch values per version
```
ovirt_system_options:
    name: "IsPortIsolationSupported"
```
That should return a dictionary with option values for all version, for example:
```
  'IsPortIsolationSupported': {
    '4.2': 'false',
    '4.3': 'false',
    '4.4': 'false',
    '4.5': 'true',
  }

```
3. Provide empty name to fetch all options
```
ovirt_system_options:
```
That should return a dictionary with all options and their values, for example
```
  'IsPortIsolationSupported': {
    '4.2': 'false',
    '4.3': 'false',
    '4.4': 'false',
    '4.5': 'true',
  },
  'DisableFenceAtStartupInSec': {
    'general': '300'
  },
...
```

PR #206, opened 2021-01-22
Add ovirt_system_option_info

mnecas:
Fixes https://github.com/oVirt/ovirt-ansible-collection/issues/204

- Add new module ovirt_system_option_info to gather information about the system options.
- We cannot list the options so there are some parameters that will not be used like fetch_nested and nested_attributes. The reason why I kept it there is that I would need to 
  - either create another doc fragment for this specific module and don't use the info fragment
  - or move all necessary arguments to this module but this could break something if we would change the main and we could forget about this

Example input:
```yaml
      ovirt.ovirt.ovirt_system_option_info:
        auth: "{{ ovirt_auth }}"
        name: ServerCPUList
        version: "4.4"
```
Example output: 
```json
    "ovirt_system_option": {
        "href": "/ovirt-engine/api/options/ServerCPUList",
        "id": "ServerCPUList",
        "name": "ServerCPUList",
        "values": [
            {
                "value": "1:Intel Nehalem Family:vmx,nx,model_Nehalem:Nehalem:x86_64; 2:Secure Intel ...",
                "version": "4.4"
            }
        ]
    }
```

### DIFF
--- a/plugins/modules/ovirt_system_option_info.py
+++ b/plugins/modules/ovirt_system_option_info.py
@@ -24,15 +24,15 @@ __metaclass__ = type
 
 DOCUMENTATION = '''
 ---
-module: ovirt_system_options_info
-short_description: Retrieve information about one oVirt/RHV system option.
+module: ovirt_system_option_info
+short_description: Retrieve information about one oVirt/RHV system options.
 version_added: "1.3.0"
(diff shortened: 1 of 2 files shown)


mwperina (approved):
LGTM

mwperina:
@arachmani

PR #214, opened 2021-02-01
ovirt_host_storage_info: no need for login to get info

mnecas:
We don't need to login to iscsi before getting all information about the host storages.
Now user can only select the host from which to get the storages and filter if iscsi or fcp.

**Examples:** 
To get all storage from host:
```
    - ovirt.ovirt.ovirt_host_storage_info:
        host: host_mixed_1
        auth: "{{ ovirt_auth }}"
```
To get only iscsi storages: 
```
    - ovirt.ovirt.ovirt_host_storage_info:
        host: host_mixed_1
        iscsi: {}
        auth: "{{ ovirt_auth }}"
```

### DIFF
--- a/plugins/modules/ovirt_host_storage_info.py
+++ b/plugins/modules/ovirt_host_storage_info.py
@@ -136,16 +136,12 @@ def main():
 
         # Get LUNs exposed from the specified target
         host_storages = host_service.storage_service().list()
-        filterred_host_storages = [host_storage for host_storage in host_storages]
         if module.params.get('iscsi') is not None:
-            filterred_host_storages = [host_storage for host_storage in host_storages
-                                       if host_storage.type == otypes.StorageType.ISCSI]
+            host_storages = list(filter(lambda x: x.type == otypes.StorageType.ISCSI, host_storages))
             if 'target' in module.params.get('iscsi'):
-                filterred_host_storages = [host_storage for host_storage in filterred_host_storages
-                                           if module.params.get('iscsi').get('target') == host_storage.logical_units[0].target]
+                host_storages = list(filter(lambda x: module.params.get('iscsi').get('target') == x.logical_units[0].target, host_storages))
         elif module.params.get('fcp') is not None:
-            filterred_host_storages = [host_storage for host_storage in host_storages
-                                       if host_storage.type == otypes.StorageType.FCP]
+            host_storages = list(filter(lambda x: x.type == otypes.StorageType.FCP, host_storages))
 
         result = dict(
             ovirt_host_storages=[
@@ -154,7 +150,7 @@ def main():
                     connection=connection,
                     fetch_nested=module.params.get('fetch_nested'),
                     attributes=module.params.get('nested_attributes'),
-                ) for c in filterred_host_storages
+                ) for c in host_storages
             ],
         )
         module.exit_json(changed=False, **result)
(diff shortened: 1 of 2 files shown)


mnecas:
@mwperina

mwperina (approved):
+1

plugins/modules/ovirt_host_storage_info.py
```diff
@@ -134,26 +142,20 @@ def main():
         # Get Host
         hosts_service = connection.system_service().hosts_service()
         host_id = get_id_by_name(hosts_service, module.params['host'])
-        storage_type = _get_storage_type(module.params)
         host_service = hosts_service.host_service(host_id)
 
-        if storage_type == 'iscsi':
+        if module.params.get('iscsi'):
```
barpavel, line 147: IMHO more readable to define before this line:
    iscsi = module.params.get('iscsi')
and use everywhere (5 places).
Can be done also for fcp for readability
    module.params.get('fcp')

plugins/modules/ovirt_host_storage_info.py
```diff
@@ -134,26 +142,20 @@ def main():
...
-        if storage_type == 'iscsi':
-            filterred_host_storages = [host_storage for host_storage in host_storages
-                                       if host_storage.type == otypes.StorageType.ISCSI]
-            if 'target' in iscsi:
-                filterred_host_storages = [host_storage for host_storage in filterred_host_storages
-                                           if iscsi.get('target') == host_storage.logical_units[0].target]
-        elif storage_type == 'fcp':
-            filterred_host_storages = [host_storage for host_storage in host_storages
-                                       if host_storage.type == otypes.StorageType.FCP]
+        if module.params.get('iscsi') is not None:
```
barpavel, line 153: I think "is not None" can be removed both here and on line 157.

plugins/modules/ovirt_host_storage_info.py
```diff
@@ -134,26 +142,20 @@ def main():
...
+            host_storages = list(filter(lambda x: x.type == otypes.StorageType.ISCSI, host_storages))
```
barpavel, line 154: I less like such names as "x" (would prefer "host_storage" instead), but it looks like this is a convention :)

barpavel:
+1

barpavel (approved):

plugins/modules/ovirt_host_storage_info.py
mnecas: This is needed for the case when we would want to list all `iscsi` without login.
If I would remove it it would not pass with `iscsi: {}` so it would not filter the StorageType.
The same goes for below.

plugins/modules/ovirt_host_storage_info.py
mnecas: Yeah, it's a convention, we use often x or if there is some short name for the var.

plugins/modules/ovirt_host_storage_info.py
mnecas: Yeah, will do

plugins/modules/ovirt_host_storage_info.py
barpavel: Ack.
You are right :)

barpavel (approved):

PR #215, opened 2021-02-01
ovirt_disk: fix transfer ending

mnecas:
Issue: after uploading the image the status of the disk was OK so it tried to attach the disk to the VM but at that moment it failed because of the unfinished transfer status.

Now we wait longer after the transfer is done and it is removed.

I think it would be best to fix this in the SDK and this could be just a temporary fix.

### DIFF
--- a/plugins/modules/ovirt_disk.py
+++ b/plugins/modules/ovirt_disk.py
@@ -370,6 +370,7 @@ from ansible.module_utils.six.moves.http_client import HTTPSConnection, Incomple
 from ansible.module_utils.six.moves.urllib.parse import urlparse
 try:
     import ovirtsdk4.types as otypes
+    import ovirtsdk4
 except ImportError:
     pass
 from ansible.module_utils.basic import AnsibleModule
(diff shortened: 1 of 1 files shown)


mnecas:
@mwperina @nirs

mwperina:
SDK is generated automatically, why should we fixed something here? If something needs to be fixed, then it should be fixed in RESTAPI and/or engine backend no?

@barpavel

mwperina (changes requested):
-1 until we finalize discussion in https://bugzilla.redhat.com/show_bug.cgi?id=1883949

plugins/modules/ovirt_disk.py
```diff
@@ -459,9 +460,13 @@ def transfer(connection, module, direction, transfer_func):
         while transfer.phase in [
             otypes.ImageTransferPhase.TRANSFERRING,
             otypes.ImageTransferPhase.FINALIZING_SUCCESS,
+            otypes.ImageTransferPhase.FINISHED_SUCCESS
         ]:
             time.sleep(module.params['poll_interval'])
-            transfer = transfer_service.get()
+            try:
```
barpavel, line 466: Looks like the code is indented to the right 1 tab too much?
Lines:
465-469
476-478

barpavel:
> SDK is generated automatically, why should we fixed something here? If something needs to be fixed, then it should be fixed in RESTAPI and/or engine backend no?

Sounds reasonable. But I prefer to hear @nirs opinion as well.

mnecas:
We have decided to wait for a fix from the storage team in https://bugzilla.redhat.com/show_bug.cgi?id=1849861

PR #218, opened 2021-02-10
roles: hosted_engine_setup: Do not reboot the host during deploy

didib:
We recently changed the host-deploy process to reboot the host after
installation. We need to revert that for hosted-engine deploy, as this
reboots the host we deploy on, thus killing the deployment.

### DIFF
--- a/automation/build-artifacts.sh
+++ b/automation/build-artifacts.sh
@@ -72,6 +72,9 @@ pip3 install rstcheck antsibull-changelog ansible-lint
 
 ansible-test sanity
 /usr/local/bin/antsibull-changelog lint
-/usr/local/bin/ansible-lint roles/* -x 204
+if ! /usr/local/bin/ansible-lint roles/* -x 204; then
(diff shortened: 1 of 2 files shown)


didib:
https://jenkins.ovirt.org/view/oVirt%20system%20tests/job/ovirt-system-tests_manual/7756/

mnecas (commented):
I'll check out that CI issue, looks like a new version of yamlint.

I don't mind the change, but not sure if it's necessary. That's a question for @arachmani

There was an issue with doing something on the host right after installation but it was fixed in the engine by https://gerrit.ovirt.org/c/ovirt-engine/+/113388

mwperina (approved):
Let's get this merged and released for QE. This is a flow we forgot about with default reboot, so we need to fix asap, otherise HE installation is completely broken

mwperina:
And let's take a look at CI separately

arachmani (approved):

PR #223, opened 2021-02-12
ovirt_auth: fix env auth dict

mnecas:
Follow up of https://github.com/oVirt/ovirt-ansible-collection/pull/211
The issue was that the `ovirt_auth` module was returning and passing `timeout` and `compress` to other modules with the `auth` option. But they were not mentioned in the updated code which forces to match the `auth` before the #211 the options were ignored. 
@mwperina @didib


mwperina (approved):
LGTM

PR #229, opened 2021-03-08
Add Grafana to hosted engine

roniezr:


kobihk (approved):

roles/hosted_engine_setup/README.md
```diff
@@ -47,6 +47,7 @@ Ansible role for deploying oVirt Hosted-Engine
 | he_debug_mode | false | If `True`, HE deployment will execute additional tasks for debug |
 | he_db_password | UNDEF | Engine database password |
 | he_dwh_db_password | UNDEF | DWH database password |
+| he_enable_grafana | False | If `True`, Grafana integration will be set up. Valid for oVirt >= 4.4. |
```
arachmani, line 50: Maybe add `on the engine VM` for clarification:
...Grafana integration will be set up on the engine VM...

arachmani (approved):
Please use `role: hosted_engine_setup:` prefix in the commit message

didib:
Please note that we deliberately disabled this, see [1][2]. Does this work for you as-is? Also, I suppose that the intention is to keep the default 'Disabled' and only optionally allow enabling. Perhaps clarify this.

[1] https://gerrit.ovirt.org/q/I48aca34f31118ee1fb03ee2ba7a5fc3c850652ee

[2] https://bugzilla.redhat.com/show_bug.cgi?id=1866780

roniezr:
didid, No, the patch does not work:( Grafana is always False, so I guess we should abandon this patch, WDYT?

didib:
> didid, No, the patch does not work:( Grafana is always False, so I guess we should abandon this patch, WDYT?

I agree. If you do need it, you'll have to patch the appliance as well, but if you do, please consider the reasoning around the original bug/patch. Thanks.

roniezr:
It decided to abandon this PR, the right procedure to enable Grafana should be implemented according to instructions listed here:
https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.4/html/administration_guide/monitoring_and_observability

PR #234, opened 2021-03-15
image_template: add external vm option

mnecas:
This PR adds:
- kvm_import
- xen_import
- vmware_import 
- host_import - hostname from which the vm should be imported
https://bugzilla.redhat.com/show_bug.cgi?id=1937408


mnecas:
Continuing in: https://github.com/oVirt/ovirt-ansible-collection/pull/304

PR #238, opened 2021-03-21
role: hosted_engine_setup: Filter VLAN devices with bad names

avivtur:
Filter VLAN devices with bad names

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1933642
Signed-off-by: Aviv Turgeman <aturgema@redhat.com>


avivtur:
Verified

roles/hosted_engine_setup/tasks/validate_vlan_name.yml
```diff
@@ -0,0 +1,23 @@
+---
+- name: Get VLAN devices parent and id as naming convention pattern
+  shell: set -eo pipefail && nmcli -g VLAN.PARENT,VLAN.ID device show {{ vlan_if.nic_if.nic }} | paste -sd '.'
+  with_items:
+    - "{{ vlan_list.results | reject('skipped') | list }}"
+  loop_control:
+    loop_var: vlan_if
+  register: vlan_pattern_list
```
mnecas, line 8: The CI fails on `301 Commands should not change things if nothing needs doing`
You can fix it by adding `changed_when: true` as in the previous `shell` task (`Collect interface types`) or recommend adding some logic to check if it changed successfully

avivtur: Thank you very much for your comment, it was certainly helpful :)
I've added the suggested fix

arachmani (commented):
When we added the team filtering we had a message only for team devices that were not supported, now we have further options, so I think it's best if we add also a fail message for each unsupported case when only this type is available.

avivtur:
Verified
Followed @arachmani comments

avivtur:
ci test please

mnecas (changes requested):
Please remove all `debug` modules or add to all of them `name`. CI currently does not check this but we are moving to ansible-lint 5.0.0 where it is required.

mnecas:
ci test

mnecas:
ci test please

mnecas (approved):
LGTM

arachmani (approved):

PR #241, opened 2021-03-27
ansible-builder: add execution-environment.yml

mnecas:
**General info**
Creating of container image
`ansible-builder build --tag ovirt --container-runtime docker -v 3`

Using the image to run the playbook with dependencies with ansible-runner.
(We need to have an ansible-runner devel version which supports --container-image)
`ansible-runner playbook --container-image=ovirt playbook_name.yml -vvv `

**Issue in the PR**
Currently, the playbook fails because it does not see the collection.

I tried to debug it by adding to the end of `execution-enviroment.yml`, I added
`RUN ls -la /usr/share/ansible/collections/ansible_collections/` we can see the installed `ovirt.ovirt` collection correctly.

But when I try to list it with the playbook below and execute it as above, the playbook does not see the directory.
```
- hosts: localhost
  tasks:
    - name: List path
      find:
        paths: /usr/share/ansible/collections/ansible_collections/
```

### DIFF
--- a/execution-environment.yml
+++ b/execution-environment.yml
@@ -1,0 +1,12 @@
+---
+version: 1
+dependencies:
+  system: bindep.txt
+
+additional_build_steps:
+  prepend: |
+    RUN pip3 install --upgrade pip setuptools
+    RUN dnf install https://resources.ovirt.org/pub/yum-repo/ovirt-release-master.rpm -y
+    RUN dnf install ovirt-ansible-collection -y
+  append:
+    - RUN ls -la /usr/share/ansible/collections/ansible_collections/
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,0 @@
-ovirt-engine-sdk-python>=4.4.10


mwperina (approved):
LGTM

mnecas:
-1 Currently not working

Hi @AlanCoding @matburt,
could you please check out the PR and maybe tell me what is wrong/if there is an issue in runner/builder.
Because it looks like ansible builder installs the collection without problem but runner can't see it in the image.

execution-environment.yml
```diff
@@ -0,0 +1,12 @@
...
+dependencies:
+  system: bindep.txt
+
+additional_build_steps:
+  prepend: |
+    RUN pip3 install --upgrade pip setuptools
+    RUN dnf install https://resources.ovirt.org/pub/yum-repo/ovirt-release-master.rpm -y
+    RUN dnf install ovirt-ansible-collection -y
+  append:
+    - RUN ls -la /usr/share/ansible/collections/ansible_collections/
```
AlanCoding, line 12: https://ansible-builder.readthedocs.io/en/latest/collection_metadata.html

Collections can put metadata of this form into the collection, but it needs to be in `meta/execution-environment.yml`. However, the bigger problem here is that only a few things are allowed to be customized by the collections themselves (as opposed to the user's definition), and prepend/append steps are not allowed.

Collections can put their own dnf requirements in a `bindep.txt` file, but that doesn't solve the main problem here - which is the need for a custom repo. See https://github.com/ansible/ansible-builder/issues/191

mnecas:
Closing, we will not do the execution environment image by ourselves.

PR #243, opened 2021-03-29
infra: fix storage target

mnecas:
The issue was that the role was trying to edit the storage connection
target when the storage domain has multiple connection targets.

This patch checks if the target which user input is in the list
of targets from the storage domain if so it will ignore it else it will edit
it.

https://bugzilla.redhat.com/show_bug.cgi?id=1943221

### DIFF
--- a/roles/infra/roles/storages/tasks/main.yml
+++ b/roles/infra/roles/storages/tasks/main.yml
@@ -91,7 +91,7 @@
     username: "{{ storages[item.0.name][item.0.storage.type].username | default(omit) }}"
     password: "{{ storages[item.0.name][item.0.storage.type].password | default(omit) }}"
     port: "{{ storages[item.0.name][item.0.storage.type].port | default(omit) }}"
-    target: "{{ storages[item.0.name][item.0.storage.type].target | default(omit) }}"
+    target: "{% if storages[item.0.name][item.0.storage.type].target not in item.0.storage.volume_group.logical_units | map(attribute='target') | list %}{{ storages[item.0.name][item.0.storage.type].target | default(omit) }}{% else %}{{ omit }}{% endif %}"
     force: true
   with_subelements:
     - "{{ sd_info.ovirt_storage_domains | default([]) }}"


mwperina (commented):
Fix looks good, but please add a description to commit message describing the issue and a fix for it

mwperina:
Pavel, could you please take a look?

@barpavel

mwperina:
Eyal, coul you please also take a look?

@shenitzky

roles/infra/roles/storages/tasks/main.yml
```diff
@@ -91,7 +91,7 @@
     username: "{{ storages[item.0.name][item.0.storage.type].username | default(omit) }}"
     password: "{{ storages[item.0.name][item.0.storage.type].password | default(omit) }}"
     port: "{{ storages[item.0.name][item.0.storage.type].port | default(omit) }}"
-    target: "{{ storages[item.0.name][item.0.storage.type].target | default(omit) }}"
+    target: "{% if storages[item.0.name][item.0.storage.type].target not in item.0.storage.volume_group.logical_units | map(attribute='target') | list %}{{ storages[item.0.name][item.0.storage.type].target | default(omit) }}{% else %}{{ omit }}{% endif %}"
```
barpavel, line 94: 1) Very long line. Is it possible to break to a few lines, so "else" & "endif" will be at separate lines and indented?
2) Maybe swap the condition: "not in" --> "in". IMHO will be more readable.
3) I'm not a big expert in Ansible, so sorry if it's a stupid question. But if I understand correctly, "default(omit)" means the "item.0.storage.type].target" is optional. What will happen with the "if storages[item.0.name][item.0.storage.type].target" check then (if it doesn't exist)? Is there a potential crash (similar to Null Pointer Exception in Java)?

barpavel:
> Eyal, coul you please also take a look?
> 
> @shenitzky

I would add Nir Soffer as well :)

barpavel:
Adding @nirs

roles/infra/roles/storages/tasks/main.yml
nirs, line 94: The commit message does not explain the issue:

    The issue was that the role was trying to edit the storage connection
    target when the storage domain has multiple connection targets.

What is the problem with editing the connection when there are multiple
connections?

    This patch checks if the target which user input is in the list
    of targets from the storage domain if so it will ignore it else it will edit it.

What do you mean by "ignoring the target"?

The task is called "Validate connections of storages" but I cannot understand
what this code is doing. I don't see any validation or editing.

What does ovirt_storage_connection does? add a connection?
What does it do when you don't specify a target? What does it do
when you specify a target?

The actual change is not unreadable:

        target: "{% if storages[item.0.name][item.0.storage.type].target not in item.0.storage.volume_group.logical_units | map(attribute='target') | list %}{{ storages[item.0.name][item.0.storage.type].target | default(omit) }}{% else %}{{ omit }}{% endif %}"

252 characters line?! and I don't understand the syntax. Can we extract this to
readable python function(s)?

Finally a comment is needed, explaning why special handling is needed for the target.

mnecas: > The commit message does not explain the issue:
> 
> ```
> The issue was that the role was trying to edit the storage connection
> target when the storage domain has multiple connection targets.
> ```
> 
> What is the problem with editing the connection when there are multiple
> connections?
> 
> ```
> This patch checks if the target which user input is in the list
> of targets from the storage domain if so it will ignore it else it will edit it.
> ```
> 
> What do you mean by "ignoring the target"?
> 
A few of your questions, Peter described in the bz https://bugzilla.redhat.com/show_bug.cgi?id=1943221 
I'll update the commit msg.

> The task is called "Validate connections of storages" but I cannot understand
> what this code is doing. I don't see any validation or editing.
Well, this is done in the module itself.
> 
> What does ovirt_storage_connection does? add a connection?
> What does it do when you don't specify a target? What does it do
> when you specify a target?

We use it because the storage domain cannot change much in the API, but the storage connection can change many parameters IP for NFS, paths etc. If we would remove it then it would remove the idempotency of the storage domain. 

> 
> The actual change is not unreadable:
> 
> ```
>     target: "{% if storages[item.0.name][item.0.storage.type].target not in item.0.storage.volume_group.logical_units | map(attribute='target') | list %}{{ storages[item.0.name][item.0.storage.type].target | default(omit) }}{% else %}{{ omit }}{% endif %}"
> ```
> 
> 252 characters line?! and I don't understand the syntax. Can we extract this to
> readable python function(s)?
> 
> Finally a comment is needed, explaning why special handling is needed for the target.

ACK the 252 characters line I'll split it into multiple lines, but no python function this would not be ansible best practice.

mnecas:
Closing.
Continuing in https://github.com/oVirt/ovirt-ansible-collection/pull/243

PR #250, opened 2021-04-12
role: hosted_engine_setup: Add an error message for FIPS on CentOS

arachmani:
Add an error message when using FIPS mode on an appliance based on CentOS
in case the FIPS mode cannot be applied.

Bug-Url: https://bugzilla.redhat.com/1924590


arachmani:
Verified

mnecas (approved):
LGTM

Issue #246, opened 2021-04-05
Give an option to fetch all nested_attributes for modules with fetch_nested


WsevoL0D:
Currently, since ansible 2.9, I think, we need to explicitly specify a nested attributes in list for nested_attributes field, for corresponding module(usually ovirt*info) to load them. The problem is that occasionally the names of desired attributes are not obvious.
Currently the ways are: 
to guess attributes names (most of the times they are pretty obvious, but not always)
check in older ansible version, which fetches all attributes by default.

I suggest to add keyword "all" to the list, as the way to tell the module to load all nested attrs (default behaviour for older versions).

That won't affect modules speed (as default behavior left as is)
Gives users a way to quickly figure out attribute names (alternative way to modify docs for each module with nested_attributes)
Gives users an elegant way to load all attributes if needed (alternative to provide all names in a list)

```yaml
    - name: "Gather full storage vm info"
      ovirt_storage_domain_info:
        auth: "{{ ovirt_auth }}"
        fetch_nested: true
        nested_attributes:
        - all
```

mnecas:
Hi @vsevolodZ,
thanks once again for the issue I'm opened to all suggestions.

I was doing some research and if we would get all `nested_attributes` it would take a lot of time even in an almost empty env.
So I tried to create some sorts of limits like getting only the first layer of nested_attributes it was faster but still not best plus some parameters could be missing.
Another idea which I had was to get all `nested_attributes` of specific attributes. Like, get all top layer info for VMs template. But did not have time to implement it and test it.

vsevolodZ:
Hi @mnecas 

Well, I don't have experience in ansible modules development, so not sure which is better from the developers perspective.

From user's point of view, given that anything other than full fetching of nested attributes breaks compatibility anyway. By break I mean that if the one fetches ovirt nested attrs in his plays he need to change yaml code when going from 2.8 to 2.9. And given that full fetch is not an option and given that users seemingly don't have a lot of issues with new logic (Personally for me it wasn't a big deal to change my ovirt projects for 2.9). I would left fetch logic as it currently is (it's minimal possible I guess). 

But I would include list off attribute names in doc (maybe one list for all ovirt_*_info modules), or alternatively provide an algorithm how to determine their names (if there is such algorithm, back to my SO question).

And separately I would consider implementing something similar to this feature request ))

mnecas:
[...]

PR #251, opened 2021-04-14
info modules: add follow parameter

mnecas:
We have multiple issues with the info modules.
1. the `get_dict_of_struct` function uses `connection.follow_link` which creates a request for each nested parameter and this slows the gathering of data. Using the `follow` in `list` speeds this process us because we leave it up to the engine and get all required data in one request.
2. We filter all nested data even if it was passed in the request. Now it shows exactly the same resp as in the API. (Could resolve issue https://github.com/oVirt/ovirt-ansible-collection/issues/246)

### DIFF
--- a/plugins/doc_fragments/ovirt_info.py
+++ b/plugins/doc_fragments/ovirt_info.py
@@ -25,14 +25,6 @@ options:
             - This parameter apply only when C(fetch_nested) is I(true).
         type: list
         elements: str
-    follows:
-      description:
-        - List of follow link names which will be gathered.
-        - This parameter replaces use of C(fetch_nested) and C(nested_attributes).
-        - If C(follows) is specified it uses over C(fetch_nested).
-      type: list
-      version_added: 1.4.0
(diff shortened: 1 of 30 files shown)


plugins/doc_fragments/ovirt_info.py
```diff
@@ -18,13 +18,22 @@ class ModuleDocFragment(object):
...
     nested_attributes:
         description:
             - Specifies list of the attributes which should be fetched from the API.
             - This parameter apply only when C(fetch_nested) is I(true).
+            - This parameter is deprecated and replaced by C(follows).
         type: list
         elements: str
+    follows:
+      description:
+        - List of follow link names which will be gathered.
```
mwperina, line 32: List of linked entities, which should be fetched along with the main entity

mwperina (changes requested):
We should add follows option to all our info modules

mwperina (approved):
Looks good to me

PR #252, opened 2021-04-19
infra: remove target from ovirt_storage_connection

mnecas:
The reason for this is that the storage domain connection can be created with multiple targets, the issue comes when there is another connection with a different target than the user has specified by the storage in the YAML file.
https://bugzilla.redhat.com/show_bug.cgi?id=1943221

### DIFF
--- a/roles/infra/roles/storages/tasks/main.yml
+++ b/roles/infra/roles/storages/tasks/main.yml
@@ -77,7 +77,7 @@
     - storages
     - storage_connections
 
-- name: Validate connections of storages
+- name: Update storage pamareters
   ovirt_storage_connection:
     auth: "{{ ovirt_auth }}"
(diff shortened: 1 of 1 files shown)


mwperina (approved):
Looks good to me, but we need a final ack from storage team about it

mwperina:
Merging to unblock releasing for oVirt 4.4.6. If the solution is not good, we will need to fix it later

roles/infra/roles/storages/tasks/main.yml
```diff
@@ -77,7 +77,7 @@
     - storages
     - storage_connections
 
-- name: Validate connections of storages
+- name: Update storage pamareters
```
barpavel, line 80: pamareters --> parameters
:)

mnecas: Fixed in: https://github.com/oVirt/ovirt-ansible-collection/pull/282

Issue #257, opened 2021-04-26
[DOCS] Add he_force_ip4 & he_force_ip6 to README


arachmani:
Clone from https://github.com/oVirt/ovirt-ansible-hosted-engine-setup/issues/121

PR #258, opened 2021-04-26
role: hosted_engine_setup: Add he_force_ip4 & he_force_ip6 to README

arachmani:
Fixes https://github.com/oVirt/ovirt-ansible-collection/issues/257


kobihk (approved):

mwperina:
Martin, shouldn't we have a changelog fragment for such change? I mean adding a new parameters to the role.
If so, then we really need some CI check ...

@mnecas

arachmani:
@mwperina I didn't add new parameters to the role, just updated the README.

mnecas:
I usually add all changelogs before the build but I will add the check to CI so every PR will have the changelog.
It would be nice to have it here because it is summed up for the build.

mwperina:
> @mwperina I didn't add new parameters to the role, just updated the README.

Right, my bad, sorry

PR #268, opened 2021-05-03
automation: add changelog check

mnecas:


mnecas:
ci test please

plugins/modules/ovirt_disk.py
```diff
@@ -859,6 +859,7 @@ def main():
                     wait_condition=lambda d: d.status == otypes.DiskStatus.OK,
                     storage_domain=otypes.StorageDomain(name=module.params['image_provider']),
                 )
+
```
mwperina, line 862: Do we need that change?

automation/build-artifacts.sh
```diff
@@ -75,3 +75,9 @@ ansible-test sanity
 /usr/local/bin/ansible-lint roles/* -x 204
 
 cd $ROOT_PATH
+
+# If PR changed something in ./plugins or ./roles it is required to have changelog
+if [[ $(git diff --quiet FETCH_HEAD ./plugins ./roles)$? -eq 1 && $(git diff --quiet FETCH_HEAD ./changelogs)$? -eq 0 ]]; then
+        echo "Please add changelog.";
+        exit 1;
```
mwperina, line 82: Does it mean that for every change in plugins or roles we will have CI complain for missing changelog?

mwperina: Is there an easy way how to figure out if there was a change adding some options/parameters or a new modules?

plugins/modules/ovirt_disk.py
mnecas: This is just a testing change so the CI would fail.

mnecas:
-1 currently not working

mwperina (approved):
OK, so let's add this warning always, so we can improve it once we have some better idea

PR #271, opened 2021-05-05
roles: hosted_engine_setup: align with ansible-lint 5.0.0

arachmani:
Fixes https://github.com/oVirt/ovirt-ansible-collection/issues/266

Signed-off-by: Asaf Rachmani <arachman@localhost.localdomain>

PR #273, opened 2021-05-09
roles: hosted_engine_setup: Add pause option before running engine-setup

arachmani:
Add the option to pause the deployment and adjust changes in the
bootstrap engine VM before running the engine-setup

Bug-Url: https://bugzilla.redhat.com/1959273

### DIFF
--- a/roles/hosted_engine_setup/tasks/bootstrap_local_vm/03_engine_initial_tasks.yml
+++ b/roles/hosted_engine_setup/tasks/bootstrap_local_vm/03_engine_initial_tasks.yml
@@ -65,8 +65,8 @@
       - name: Allow the user to connect to the bootstrap engine and change configuration
         debug:
           msg: >-
-            You can now connect to the bootstrap engine using the temporary IP address -
-            {{ hostvars[he_ansible_host_name]['local_vm_ip']['stdout_lines'][0] }}
+            You can now connect from this host to the bootstrap engine VM using ssh as root
+            and the temporary IP address - {{ hostvars[he_ansible_host_name]['local_vm_ip']['stdout_lines'][0] }}
       - include_tasks: ../pause_execution.yml
     when: he_pause_before_engine_setup|bool
   - name: Restore a backup
--- a/roles/hosted_engine_setup/tasks/bootstrap_local_vm/03_engine_initial_tasks.yml
+++ b/roles/hosted_engine_setup/tasks/bootstrap_local_vm/03_engine_initial_tasks.yml
@@ -60,6 +60,15 @@
     include_tasks: "{{ item }}"
     with_fileglob: "hooks/enginevm_before_engine_setup/*.yml"
     register: include_before_engine_setup_results
+  - name: Pause the execution to allow the user to configure the bootstrap engine VM
+    block:
+      - name: Allow the user to connect to the bootstrap engine and change configuration
+        debug:
+          msg: >-
+            You can now connect to the bootstrap engine using the temporary IP address -
+            {{ hostvars[he_ansible_host_name]['local_vm_ip']['stdout_lines'][0] }}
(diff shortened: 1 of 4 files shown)


roles/hosted_engine_setup/tasks/bootstrap_local_vm/03_engine_initial_tasks.yml
```diff
@@ -54,16 +54,24 @@
       - "OVESETUP_DWH_DB/password=str:{{ he_dwh_db_password }}"
     when: he_dwh_db_password is defined
   - name: Import OpenSCAP task
-    import_tasks: apply_openscap_profile.yml
+    import_tasks: ../apply_openscap_profile.yml
```
didib, line 57: Why do we need to import from "../"? And is it only in this file? Perhaps worth a separate patch.

roles/hosted_engine_setup/tasks/bootstrap_local_vm/03_engine_initial_tasks.yml
```diff
@@ -54,16 +54,24 @@
...
     include_tasks: "{{ item }}"
     with_fileglob: "hooks/enginevm_before_engine_setup/*.yml"
     register: include_before_engine_setup_results
-  - debug: var=include_before_engine_setup_results
+  - name: Pause the execution to allow the user to configure the bootstrap engine VM
+    block:
+      - name: Allow the user to connect to the bootstrap engine and change configuration
+        debug:
+          msg: >-
+            You can now connect to the bootstrap engine using the temporary IP address -
```
didib, line 68: Perhaps 'using ssh as root, from this host, to the temporary...'. I suppose it's only from this host, right?

didib:
Also, perhaps add a Bug-Url: - perhaps https://bugzilla.redhat.com/1956413 .

Also, perhaps pause (optionally, but true by default) after engine-setup, if it fails, like he_pause_after_failed_add_host .

roles/hosted_engine_setup/tasks/bootstrap_local_vm/03_engine_initial_tasks.yml
arachmani: This is handled in https://github.com/oVirt/ovirt-ansible-collection/pull/271, so we need to rebase it.

didib: I do not see that you rebased. Not sure about your exact plans nor whether github makes it easy to rebase a PR on top of another PR.

didib:
> > Also, perhaps add a Bug-Url: - perhaps https://bugzilla.redhat.com/1956413 .
> > Also, perhaps pause (optionally, but true by default) after engine-setup, if it fails, like he_pause_after_failed_add_host .
> 
> I've opened a bug and added the Bug-Url, I prefer to add the `he_pause_after_failed...` in a different PR.

I see Bug-Url only in the PR, not on any specific patch - and AFAICT, won't make it to the git log.

arachmani:
> I see Bug-Url only in the PR, not on any specific patch - and AFAICT, won't make it to the git log.

Done, added to the last commit.

didib:
For some reason it has "Bug-Url: Bug-Url:", twice. Not sure our tooling will parse this correctly.
I'd also personally just squash everything to a single patch, I do not see much value in merging these multiple patches, but don't mind that much.

arachmani:
> For some reason it has "Bug-Url: Bug-Url:", twice. Not sure our tooling will parse this correctly.
> I'd also personally just squash everything to a single patch, I do not see much value in merging these multiple patches, but don't mind that much.

Thanks, fixed the double Bug-Url.
We will squash it to a single commit when merging.

PR #279, opened 2021-05-27
CI: switch from ansible to ansible-core

mnecas:
CI: switch from ansible to ansible-core


mwperina:
We need to maintain 2.9 compatibility, so I'd prefer to have both ansible version tested

mnecas:
I'm not sure if it is possible when I tried to install both 2.11 and 2.9 I got `Cannot install ansible-core with a pre-existing ansible==2.9.21`. 
And this change only affects the sanity tests where the 2.11 will have just a few more tests.
Am I right @thedoubl3j?

thedoubl3j:
You cannot install both on top of each other, they would have to be in separate testing environments. Or you could test one, remove it and install the second one. 

And correct @mnecas this is only for sanity testing. Backwards compat will still be met with 2.9 sanity suite, 2.11 just contains some new tests (and some might have been removed, I haven't checked the changelog in while). If you are passing 2.11 sanity suite, you all should have no problem passing 2.9.

automation/build-artifacts.ansible.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -xe
+
+dnf install ansible ansible-test
+
+export ANSIBLE_EXEC_PREFIX="/bin"
```
mwperina, line 5: Hmm, isn't ansible installed by default as `/usr/bin/ansible`?

mwperina (changes requested):
There are bunch of changes around default values? Why so many changes are needed?

Also do we have a consistent approach around default values? Personally I'd put default value into ansible doc only if the default value is set within ansible code/. If there is a default value set by ovirt-engine, then I'd mention this in ansible code, because it can be change and we would have inconsistencies. Could you please check if above changes are aligned with this approach?

mwperina (approved):
+1

mnecas:
ci test please

PR #286, opened 2021-06-03
DR: change conf paths

mnecas:
Issue: When installing only from RPM the paths which were in the `dr.conf` were missing so after the DR started it created the paths and it blocked the usage of the collection because the `~/.ansible` path has higher priority than the `/usr/share/ansible`
https://bugzilla.redhat.com/show_bug.cgi?id=1965228


barpavel (approved):
Looks good to me.

PR #295, opened 2021-06-13
roles: hosted_engine_setup: Remove cloud-init configuration

arachmani:
Remove cloud-init configuration from the engine VM

Bug-Url: https://bugzilla.redhat.com/1965456

### DIFF
--- a/roles/hosted_engine_setup/tasks/create_target_vm/02_engine_vm_configuration.yml
+++ b/roles/hosted_engine_setup/tasks/create_target_vm/02_engine_vm_configuration.yml
@@ -86,3 +86,6 @@
       regexp: "^\\s*PermitRootLogin"
       line: "PermitRootLogin {{ he_root_ssh_access }}"
       state: present
+  - name: Clean cloud-init configuration
+    include_tasks: ../clean_cloud_init_config.yml
+    ignore_errors: true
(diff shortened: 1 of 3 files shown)


roles/hosted_engine_setup/tasks/create_target_vm/02_engine_vm_configuration.yml
```diff
@@ -86,3 +86,6 @@
       regexp: "^\\s*PermitRootLogin"
       line: "PermitRootLogin {{ he_root_ssh_access }}"
       state: present
+  - name: Clean cloud-init configuration
+    include_tasks: ../clean_cloud_init_config.yml
+    ignore_errors: true
```
sandrobonazzola, line 91: why do we need `ignore_errors: true` here?

sandrobonazzola (commented):
just a question inline, the PR looks good to me otherwise.

roles/hosted_engine_setup/tasks/create_target_vm/02_engine_vm_configuration.yml
arachmani: We don't really need it, I'll remove that.

sandrobonazzola (approved):
LGTM

arachmani:
Verified

PR #301, opened 2021-06-21
ovirt_vm: add default to check_placement_policy

mnecas:
https://bugzilla.redhat.com/show_bug.cgi?id=1973640


arachmani (approved):
Verified with HE deployment

PR #302, opened 2021-06-21
roles: hosted_engine_setup: Use FQDN instead of IP for ansible host

arachmani:
`delegate_to: "{{ groups.engine[0] }}"` currently tries to reach the engine VM using IP address.
Reaching the bootstrap VM will succeed but, the operation will fail once the local VM is present,
since it uses a different IP address.
Using FQDN instead of an IP address should fix the issue.

Bug-Url: https://bugzilla.redhat.com/1973640


arachmani:
Fixed in https://github.com/oVirt/ovirt-ansible-collection/pull/305

PR #303, opened 2021-06-22
Ignore unreachable in full_execution "Sync on engine machine"

didib:
Bug-Url: https://bugzilla.redhat.com/1973640

### DIFF
--- a/changelogs/fragments/303-HE-ignore-unreachable-full-execution.yml
+++ b/changelogs/fragments/303-HE-ignore-unreachable-full-execution.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - hosted_engine_setup - Ignore unreachable in full_execution "Sync on engine machine" (https://github.com/oVirt/ovirt-ansible-collection/pull/303)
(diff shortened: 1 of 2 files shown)


didib:
Verified by running with full_execution.

arachmani:
@didib, I think `Sync on engine machine` task at this point will always fail (unless bootstrap VM and local VM hold the same IP address), so why not remove this task?

didib:
> @didib, I think `Sync on engine machine` task at this point will always fail (unless bootstrap VM and local VM hold the same IP address), so why not remove this task?

I am not completely certain it will always fail. At the time, I simply added it right before every "Fetch logs from the engine VM" - in this case, the one done in final_clean.yml. Anyway, I don't mind removing, if you are certain it's never called.

didib:
Anyway, I'd rather merge now and then think about removing. The bug is blocking QE, as they use full_execution.

didib:
See #305 which is an alternative, preferred by Asaf. Personally I don't mind that much.

PR #304, opened 2021-06-22
ovirt_template: add ova import

mnecas:
https://bugzilla.redhat.com/show_bug.cgi?id=1937408


plugins/modules/ovirt_template.py
```diff
@@ -845,6 +873,49 @@ def _get_vnic_profile_mappings(module):
...
+        )
+    )
+
+    # Wait until event with code 1158 for our template:
+    templates_service = connection.system_service().templates_service()
+    wait(
+        service=templates_service.template_service(imported_template.template.id),
+        condition=lambda tmp: len(events_service.list(
+            from_=int(last_event.id),
+            search='type=1158 and message=*%s*' % tmp.name,
```
mnecas, line 910: Currently, there is an issue in event search with templates the `message=*%s*` is a workaround for it. 
Opened an issue on it:
https://bugzilla.redhat.com/show_bug.cgi?id=1974656

plugins/modules/ovirt_template.py
```diff
@@ -845,6 +873,49 @@ def _get_vnic_profile_mappings(module):
...
+            cluster=otypes.Cluster(
+                name=module.params['cluster'],
+            ) if module.params['cluster'] else None,
+            storage_domain=otypes.StorageDomain(
+                name=external_template.get('storage_domain'),
+            ) if external_template.get('storage_domain') else None,
+            host=otypes.Host(
+                name=external_template.get('host'),
+            ) if external_template.get('host') else None,
+        )
```
mnecas, line 901: There was another parameter `clone` added in the API but the python SDK was not yet released with it so after the merge, I'll create an issue on it to add the parameter.
https://github.com/oVirt/ovirt-engine-api-model/blob/master/src/main/java/types/ExternalTemplateImport.java#L114-L129

mwperina (approved):
+1

PR #311, opened 2021-06-24
role: hosted_engine_setup: Fix engine vm add_host for the target machine

didib:
Allow custom hooks that want to access the engine VM as an ansible host
to work also after it is started as the target public machine.

### DIFF
--- a/roles/hosted_engine_setup/tasks/create_target_vm/01_create_target_hosted_engine_vm.yml
+++ b/roles/hosted_engine_setup/tasks/create_target_vm/01_create_target_hosted_engine_vm.yml
@@ -7,6 +7,9 @@
     environment: "{{ he_cmd_lang }}"
     register: local_vm_ip
     changed_when: true
+  - name: Set the name for add_host
+    set_fact:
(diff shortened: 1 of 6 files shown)


roles/hosted_engine_setup/tasks/create_target_vm/01_create_target_hosted_engine_vm.yml
```diff
@@ -7,6 +7,9 @@
     environment: "{{ he_cmd_lang }}"
     register: local_vm_ip
     changed_when: true
+    - name: Set the name for add_host
+      set_fact:
+        he_fqdn_ansible_host: "{{ local_vm_ip.stdout_lines[0] }}"
```
kobihk, line 12: looks like wrong indentation

didib: Fixed, thanks

kobihk (approved):
verified with our Jenkins job

didib:
Added changelog

arachmani (approved):

PR #314, opened 2021-06-29
roles: hosted_engine_setup: remove duplicate tasks

arachmani:
When using `hosted-engine --deploy`, "Sync on engine machine" and
"Final clean" tasks run twice, once at the end of "Creating Target VM" stage
and right after that in "Cleaning temporary resources" stage.
This PR removes these tasks from "Creating Target VM" stage,
therefore they will be executed only in the "Final clean" stage.

Bug-Url: https://bugzilla.redhat.com/1977486


didib:
OK for me. Did you verify that you manage to get local vm logs on both success and failure?

arachmani:
Verified

PR #315, opened 2021-07-01
roles: hosted_engine_setup: Use forward network during an IPv6 deployment

arachmani:
Use forward network instead of isolated network
during an IPv6 deployment

Bug-Url: https://bugzilla.redhat.com/1947709

### DIFF
--- a/roles/hosted_engine_setup/tasks/alter_libvirt_default_net_configuration.yml
+++ b/roles/hosted_engine_setup/tasks/alter_libvirt_default_net_configuration.yml
@@ -14,7 +14,7 @@
         xpath: /network/ip
         state: absent
       register: editednet_noipv4
-    - name: Configure libvirt default network as an forward network
+    - name: Configure libvirt default network as a forward network
       xml:
         xmlstring: "{{ editednet_noipv4.xmlstring }}"
         xpath: /network/forward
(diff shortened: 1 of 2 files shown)


roles/hosted_engine_setup/tasks/alter_libvirt_default_net_configuration.yml
```diff
@@ -6,21 +6,37 @@
...
+    - name: Accept IPv6 Router Advertisements
+      shell: echo 2 > /proc/sys/net/ipv6/conf/ovirtmgmt/accept_ra
     - name: Remove IPv4 configuration
       xml:
         xmlstring: "{{ default_net_xml.get_xml }}"
         xpath: /network/ip
         state: absent
       register: editednet_noipv4
-    - name: Configure it as an isolated network
+    - name: Configure libvirt default network as an forward network
```
didib, line 17: nit: "_a_ foward" (not "an"). Typo was there before your patch, but you can fix it :-)

didib (approved):
I don't know much about libvirt and IPv6. Patch looks ok to me, if it works.

I wonder if we should have such a difference between IPv4 and IPv6, if we now essentially do the same. NVM.

sandrobonazzola (approved):

roles/hosted_engine_setup/tasks/alter_libvirt_default_net_configuration.yml
sandrobonazzola: 7b0a59f includes the fix for the typo

PR #317, opened 2021-07-14
Upgrading RHVH hyperconverged host from Admin portal

rchikatw:
fails for first host

Upgrade is failing because ovit-ha-agent is trying to connect vdsm,
and if connection fails, it exits with a non-zero exit code.

after it fails systemd tries to restart the agent, and sincie the
agent wants vdsmd, it also starts vdsmd. This means that if you do:
systemctl stop vdmsd
and ovirt-ha-agent is in the middle of restart loop,
vdsmd service will start again. This is an impossible situation
so hence just before upgrade stopping the agent and
broker service.

Bug-Url: https://bugzilla.redhat.com/1979539
Signed-off-by: Ritesh Chikatwar <rchikatw@redhat.com>


rchikatw:
ci add to whitelist

roles/cluster_upgrade/tasks/upgrade.yml
```diff
@@ -84,6 +100,15 @@
     reboot_after_upgrade: "{{ reboot_after_upgrade }}"
     timeout: "{{ upgrade_timeout }}"
 
+- name: Start ovirt-ha-agent service
+  service:
+    name: ovirt-ha-agent
```
mnecas, line 105: Shouldn't we also start the `ovirt-ha-broker`?

rchikatw: No because ha-agent will start it. because ha-agent needs ha-broker

rchikatw: i verified it locally.

dangel101:
LGTM

mnecas (changes requested):
Noticed that there is a similar patch in the host upgrade [1], because we are calling host upgrade from cluster upgrade the code is not required here.
Could you please retry it locally without this change and only in host upgrade?
[1] https://gerrit.ovirt.org/#/c/ovirt-engine/+/115636/

mwperina (changes requested):
-2

That should fixed inside host deploy and not in cluster upgrade role

rchikatw:
True this is not required because host upgrade is being called internally. Hence closing this PR.

Issue #322, opened 2021-07-31
Unable to change disk shareable status


slowerogue:
##### SUMMARY
I am trying to modify existing disk in rhv to shareable/non shareable

##### COMPONENT NAME
ovirt_disk

##### STEPS TO REPRODUCE

```yaml
    - ovirt_disk:
        auth: "{{ ovirt_auth }}"
        id:  63e67aba-43ec-45c1-b495-16053ca59fd4
        state: present
        shareable: false
        storage_domain: some_domain_name

```

##### EXPECTED RESULTS
disk shareable status can be changed

##### ACTUAL RESULTS

```paste below
ok: [localhost] => {
    "changed": false,
    "diff": {
        "after": {},
        "before": {}
    },
    "disk": {
        "actual_size": 1073741824,
        "alias": "sharetestvm",
        "backup": "none",
        "content_type": "data",
        "description": "",
        "disk_profile": {
            "href": "/ovirt-engine/api/diskprofiles/55bb175d-488c-4c12-95ce-d913477c3333",
            "id": "55bb175d-488c-4c12-95ce-d913477c3333"
        },
        "disk_snapshots": [],
        "format": "raw",
        "href": "/ovirt-engine/api/disks/63e67aba-43ec-45c1-b495-16053ca59fd4",
        "id": "63e67aba-43ec-45c1-b495-16053ca59fd4",
        "image_id": "64808947-2e01-48e1-95f4-fa2843ef1a32",
        "name": "sharetestvm",
        "permissions": [],
        "propagate_errors": false,
        "provisioned_size": 1073741824,
        "quota": {
            "href": "/ovirt-engine/api/datacenters/a078b7d8-0154-11eb-ba6e-00163e65129f/quotas/a5de1178-0154-11eb-a616-00163e65129f",
            "id": "a5de1178-0154-11eb-a616-00163e65129f"
        },
        "shareable": true,
        "sparse": false,
        "statistics": [],
        "status": "ok",
        "storage_domains": [
            {
                "href": "/ovirt-engine/api/storagedomains/1b43fff0-414d-4255-ace8-90928272ebb8",
                "id": "1b43fff0-414d-4255-ace8-90928272ebb8"
            }
        ],
        "storage_type": "image",
        "wipe_after_delete": false
    },
    "id": "63e67aba-43ec-45c1-b495-16053ca59fd4",
    "invocation": {
        "module_args": {
            "activate": null,
            "auth": {
            [REDACTED]
            },
[...]

PR #323, opened 2021-08-02
ovirt_disk: fix update_check with no vm

mnecas:
Fixes: https://github.com/oVirt/ovirt-ansible-collection/issues/322

The disk which had no VM has not checked the params for the update.


mnecas:
ci test please

PR #324, opened 2021-08-02
engine_setup: wait after engine-config reboot

mnecas:
Issue: when immediately after engine_setup someone tries to access the web server it is not available because it is starting


mwperina (approved):
+1

PR #332, opened 2021-08-24
ovirt_auth: fix token no_log

mnecas:
Issue: when passing the already preheated token the ansible filters the value even for other tasks so the auth immediately fails.


mnecas:
ci test

mnecas:
Verified by running playbook with ovirt_auth where we pass token.

dangel101:
LGTM

sgratch:
LGTM

PR #346, opened 2021-09-14
replace redhat_subscription and rhsm_repository with command

mnecas:
The downstream ansible execution environment does not contain any community collection.
(The `redhat_subscription` and `rhsm_repository` are in the community collections)
We need to switch to some other solution from the modules to make it compatible with the execution environment.

### DIFF
--- a/roles/repositories/tasks/rh-subscription.yml
+++ b/roles/repositories/tasks/rh-subscription.yml
@@ -25,6 +25,7 @@
     --password {{ ovirt_repositories_rh_password }}
     {% if ovirt_repositories_force_register is defined %} --force {% endif %}
     {% if ovirt_repositories_rhsm_server_hostname is defined %} --serverurl {{ ovirt_repositories_rhsm_server_hostname }} {% endif %}
+  changed_when: false
 
 - include_tasks: search-pool-id.yml
   with_items:
--- a/roles/repositories/tasks/rh-subscription.yml
+++ b/roles/repositories/tasks/rh-subscription.yml
@@ -18,40 +18,27 @@
     name: subscription-manager
     state: present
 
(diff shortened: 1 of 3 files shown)


roles/repositories/tasks/rh-subscription.yml
```diff
@@ -18,40 +18,27 @@
...
-  redhat_subscription:
-    state: present
-    force_register: "{{ ovirt_repositories_force_register }}"
-    username: "{{ ovirt_repositories_rh_username | mandatory }}"
-    password: "{{ ovirt_repositories_rh_password | mandatory }}"
-    pool_ids: "{{ ovirt_repositories_pool_ids }}"
-    server_hostname: "{{ ovirt_repositories_rhsm_server_hostname | default(omit) }}"
-  when: ovirt_repositories_pool_ids is defined
+- name: Register to subscription manager
+  command: subscription-manager register --username {{ ovirt_repositories_rh_username }} --password {{ ovirt_repositories_rh_password }} {% if ovirt_repositories_force_register is defined %} --force {% endif %} {% if ovirt_repositories_rhsm_server_hostname is defined %} --serverurl {{ovirt_repositories_rhsm_server_hostname}} {% endif %}
```
mwperina, line 22: Can this line be split to improve readability?

mwperina (changes requested):

kobihk:
Hi @mnecas, 
Could you please elaborate in the description why we need to replace redhat_subscription and rhsm_repository with command?
I just wondering if we should expect side effects where we using these ansible-modules

mnecas:
@kobihk updated comment

kobihk:
> @kobihk updated comment

@mnecas Thank you!!

PR #355, opened 2021-10-06
info modules: enable follow parameter

mnecas:


PKehnel:
Any update when this is getting merged?

PR #357, opened 2021-10-07
ovirt_remove_stale_lun: Allow user to remove multiple LUNs

nijinashok:
Currently, if a user has to remove multiple LUNs, they have to run
the playbook individually for each LUNs. This is not that user
friendly if the user has to remove many LUNs. This commit provides
option to pass multiple LUNs.


mnecas:
@vjuranek

PR #361, opened 2021-10-15
ovirt_disk: fix vm names search and allow attaching to vm

mnecas:


mwperina:
Please add commit message, it's hard to review when it's not clear what should be fixed

mwperina (approved):
+1

plugins/modules/ovirt_disk.py
```diff
@@ -697,7 +697,7 @@ def searchable_attributes(module):
     attributes = {
         'name': module.params.get('name'),
         'Storage.name': module.params.get('storage_domain'),
-        'vm_names': module.params.get('vm_name'),
+        'vm_names': module.params.get('vm_name') if module.params.get('state') != 'attached' else None,
```
barpavel, line 700: Minor: in all places "module.params['state']" syntax is used.

barpavel (approved):

plugins/modules/ovirt_disk.py
```diff
@@ -765,7 +765,7 @@ def searchable_attributes(module):
...
```
barpavel, line 768: Now after looking at it again, have 1 question, should this `vm_name` filtering by `state` should be be done here on when used according to the flow?
I mean this `searchable_attributes()` looks like a pretty generic method, that has a potential to be used in multiple places.

mwperina: If it's worth, please do that in followup patch

PR #371, opened 2021-11-01
ovirt_host: fix failed_state_after_reinstall condition

mnecas:
failed_state_after_reinstall uses the host, so if the host was missing it automatically failed


mwperina (approved):
+1

PR #373, opened 2021-11-04
add virtio_scsi_multi_queues parameter to ovirt_vm

saifabusaleh:
Add virtio_scsi_multi_queues parameter to ovirt_vm as part of:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1973251

Requires review from: @mnecas  @ahadas @liranr23

### DIFF
--- a/plugins/modules/ovirt_vm.py
+++ b/plugins/modules/ovirt_vm.py
@@ -914,6 +914,14 @@ options:
             - "If `true`, each virtual interface will get the optimal number of queues, depending on the available virtual Cpus."
         type: bool
         version_added: 1.7.0
+    virtio_scsi_multi_queues:
+        description:
+            - "Number of queues for a Virtio-SCSI controller, possible values:
+              -1 - Indicates that the queues will be automatically set.
+               0 - Indicates that the Virtio SCSI multi-queue will be disabled.
+              >0 - Number of Virtio SCSI queues to use by virtual machine."
+        type: int
+        version_added: 1.7.0
 
 notes:
     - If VM is in I(UNASSIGNED) or I(UNKNOWN) state before any operation, the module will fail.
@@ -1577,6 +1585,7 @@ class VmsModule(BaseModule):
                 enabled=self.param('virtio_scsi_enabled'),
             ) if self.param('virtio_scsi_enabled') is not None else None,
             multi_queues_enabled=self.param('multi_queues_enabled'),
+            virtio_scsi_multi_queues=self.param('virtio_scsi_multi_queues'),
             os=otypes.OperatingSystem(
                 type=self.param('operating_system'),
                 boot=otypes.Boot(
@@ -1776,6 +1785,7 @@ class VmsModule(BaseModule):
             equal(self.param('numa_tune_mode'), str(entity.numa_tune_mode)) and
             equal(self.param('virtio_scsi_enabled'), entity.virtio_scsi.enabled) and
             equal(self.param('multi_queues_enabled'), entity.multi_queues_enabled) and
+            equal(self.param('virtio_scsi_multi_queues'), entity.virtio_scsi_multi_queues) and
             equal(self.param('rng_device'), str(entity.rng_device.source) if entity.rng_device else None) and
             equal(provided_vm_display.get('monitors'), getattr(vm_display, 'monitors', None)) and
             equal(provided_vm_display.get('copy_paste_enabled'), getattr(vm_display, 'copy_paste_enabled', None)) and
@@ -2615,6 +2625,7 @@ def main():
         next_run=dict(type='bool'),
         virtio_scsi_enabled=dict(type='bool'),
         multi_queues_enabled=dict(type='bool'),
+        virtio_scsi_multi_queues=dict(type='int'),
         snapshot_name=dict(type='str'),
         snapshot_vm=dict(type='str'),
     )
--- a/changelogs/fragments/371-ovirt_vm-add-virtio_scsi_multi_queues.yml
+++ b/changelogs/fragments/371-ovirt_vm-add-virtio_scsi_multi_queues.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - ovirt_vm - Add virtio_scsi_multi_queues (https://github.com/oVirt/ovirt-ansible-collection/pull/373).


plugins/modules/ovirt_vm.py
```diff
@@ -911,9 +911,14 @@
         version_added: 1.7.0
     multi_queues_enabled:
         description:
-            - "If `true`, each virtual interface will get the optimal number of queues, depending on the available virtual Cpus."
+            - "If `true` and multi_queues parameter is not specified, each virtual interface will get the optimal number of queues, depending on the available virtual Cpus."
```
mnecas, line 914: E501: line too long (173 > 160 characters)

mnecas (changes requested):
Please add a new fragment to changelogs that contain the change info.
The code LGTM just a small nitpick from ci.

plugins/modules/ovirt_vm.py
```diff
@@ -911,9 +911,15 @@
...
     multi_queues_enabled:
         description:
-            - "If `true`, each virtual interface will get the optimal number of queues, depending on the available virtual Cpus."
+            - "If `true` and virtio_scsi_multi_queues parameter is not specified, each virtual interface will get the optimal number of queues,
+              depending on the available virtual Cpus."
         type: bool
         version_added: 1.7.0
+    virtio_scsi_multi_queues:
+        description:
+            - "Number of Virtio SCSI queues used by virtual machine."
```
ahadas, line 920: if you're adding the enable/disable option for virtio scsi multi-queue, this should explain the relation between them (e.g., what if the boolean flag is set to enabled while this is set to 0, whether this argument can be set to -1 - such things)

ahadas (changes requested):

plugins/modules/ovirt_vm.py
saifabusaleh: Done, what you think now?

plugins/modules/ovirt_vm.py
```diff
@@ -911,9 +911,15 @@
         version_added: 1.7.0
     multi_queues_enabled:
         description:
-            - "If `true`, each virtual interface will get the optimal number of queues, depending on the available virtual Cpus."
+            - "If `true` and virtio_scsi_multi_queues parameter is not specified, each virtual interface will get the optimal number of queues,
```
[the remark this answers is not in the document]
saifabusaleh: You are right, fixed

saifabusaleh:
Tested locally the changes with the new parameter and its working

- Verified locally

mnecas:
@ahadas okay with you to merge?

mwperina (approved):
+1

plugins/modules/ovirt_vm.py
```diff
@@ -906,14 +906,23 @@
         type: str
     virtio_scsi_enabled:
         description:
-            - "Enable Virtio SCSI support."
+            - "If I(true), the Virtio-SCSI devices will obtain a number of multiple queues depending on
+               the available virtual Cpus and disks, or according to the specified virtio_scsi_multi_queues."
```
ahadas, line 910: AFAIR we don't enable the virtio-scsi multi queue by default when enabling virtio-scsi (adding the virtio-scsi controller) - did you check it is indeed enabled by default? if it's not the case, we shouldn't change this comment

plugins/modules/ovirt_vm.py
```diff
@@ -906,14 +906,23 @@
...
     multi_queues_enabled:
         description:
             - "If `true`, each virtual interface will get the optimal number of queues, depending on the available virtual Cpus."
         type: bool
         version_added: 1.7.0
+    virtio_scsi_multi_queues:
+        description:
+            - "Number of queues for a Virtio-SCSI contoller, possible values:
+              -1 - Indicates that the queues will be automatically set, see virtio_scsi_enabled
+               0 - Indicates that the Virtio SCSI will be disabled (virtio_scsi_enabled false)
```
ahadas, line 922: it doesn't indicate that virtio-scsi will be disabled but rather that virtio-scsi multi-queue will be disabled (when virtio_scsi_enabled is set, we add the virtio-scsi controller and so we can add virtio-scsi disks)

ahadas (changes requested):
I'm ok with not exposing the virtio-scsi multi-queue enabled flag but the documentation introduced in this PR should change

plugins/modules/ovirt_vm.py
saifabusaleh: My bad, actually I got confused twice with those values

now I thought virtio_scsi_enabled is actually virtio_scsi_multi_queues_enabled, because of that I changed its description

I didn't realize until now that there is no virtio_scsi_multi_queues_enabled, since this value was not added by me (it already existed before my change in API/Server)

are we ok with not exposing virtio_scsi_multi_queues_enabled flag?
users indeed can do everything they want with virtio_scsi_multi_queues parameter but this will create mismatch between ansible-collection and python-sdk/API

what do you think?

plugins/modules/ovirt_vm.py
saifabusaleh: Fixed

plugins/modules/ovirt_vm.py
ahadas: I think it's ok, we can also deprecate virtio_scsi_multi_queues_enabled if we didn't do that already 
CC @liranr23

plugins/modules/ovirt_vm.py
```diff
@@ -914,6 +914,14 @@
             - "If `true`, each virtual interface will get the optimal number of queues, depending on the available virtual Cpus."
         type: bool
         version_added: 1.7.0
+    virtio_scsi_multi_queues:
+        description:
+            - "Number of queues for a Virtio-SCSI contoller, possible values:
+              -1 - Indicates that the queues will be automatically set, see virtio_scsi_enabled
```
ahadas, line 920: "see virtio_scsi_enabled" should be dropped

saifabusaleh: Fixed

ahadas (approved):

plugins/modules/ovirt_vm.py
liranr23: we didn't. we could do that since we introduced the number of multi queues. but in order to really do that some work need to be done on engine side (and api-model). basically it means to remove our consideration if multi queues are enabled or not. while considering the enabled to act as before.
i think we actually avoided it on the bug.

ahadas: > we didn't. we could do that since we introduced the number of multi queues. but in order to really do that some work need to be done on engine side (and api-model). basically it means to remove our consideration if multi queues are enabled or not. while considering the enabled to act as before. i think we actually avoided it on the bug.

didn't we do this already as part of https://gerrit.ovirt.org/c/ovirt-engine/+/115770/?

liranr23: i don't think so, as in BackendVmsResource we also have validation to have it enabled in order to set the number value.

PR #377, opened 2021-11-05
Automation: change collection version to 2.0.0 beta

mnecas:
We need to change the version to 2.0.0 because of the depreciation of manageiq

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
-VERSION="1.7.0"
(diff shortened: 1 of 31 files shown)


mnecas:
Currently, the ci fails because in info modules are parameters that should be deprecated in 2.0.0
We have two options either remove those parameters or deprecate them at 3.0.0

mnecas:
ci test please

mwperina (approved):
+1

PR #384, opened 2021-11-22
manageiq: add deprecation info

mnecas:


roles/manageiq/README.md
```diff
@@ -1,5 +1,6 @@
 Deploy ManageIQ in oVirt
 ==================================================
+## This role is going to be deprecated in 2.0.0
```
mwperina, line 3: This role is going to be removed in 2.0.0

mwperina (changes requested):

mwperina (approved):
+1

mwperina (approved):
+1

PR #389, opened 2021-11-23
Backport: ovirt_remove_stale_lun: Retry "multipath -f" while removing the LUNs

nijinashok:
There are possiblities that the vdsm might be holding the LUN
temporarily while we run "multipath -f". In these cases, the flushing
fails with error "map in use". Retry multipath -f (6 * 5 seconds) to
workaround this race condition.


vjuranek (approved):

mwperina (approved):
+1

mnecas (approved):

PR #394, opened 2021-11-29
roles: hosted_engine_setup: Remove SPICE from graphic protocols

arachmani:
Bug-Url: https://bugzilla.redhat.com/2026770


mwperina (approved):
+1

mwperina (approved):
+1

mwperina:
Arik, could you please also take a look?

@ahadas

ahadas (approved):
I see QXL is not specified so removing SPICE and relying on libvirt & ovirt-engine to set a supported video device seems good to me

didib (approved):
As I wrote elsewhere (not sure where, sorry), from functionality POV I think we can completely get rid of the graphical console and keep only text/serial.

PR #398, opened 2021-12-02
Remove .gitignore from changelog/fragments

mwperina:
Remove .gitignore from changelog/fragments directory to pass
ansible-core 2.12 sanity checks

Signed-off-by: Martin Perina <mperina@redhat.com>

PR #399, opened 2021-12-06
Fix ansible-test ci

mnecas:
We need to ignore the ansible-test import, looks like some issue in the ansible test code.

### DIFF
--- a/automation/build.sh
+++ b/automation/build.sh
@@ -70,8 +70,10 @@ cd $COLLECTION_DIR
 
 pip3 install rstcheck antsibull-changelog "ansible-lint<5.0.0"
 
+# The sanity import test failed with error. (https://github.com/ansible/ansible/issues/76473)
 ansible-test sanity --skip-test import
 antsibull-changelog lint
+# 204 - lines should be no longer than 160 chars
 ansible-lint roles/* -x 204
 
 cd $ROOT_PATH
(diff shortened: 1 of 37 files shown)


automation/build.sh
```diff
@@ -70,9 +70,9 @@ cd $COLLECTION_DIR
 
 pip3 install rstcheck antsibull-changelog "ansible-lint<5.0.0"
 
-ansible-test sanity
-/usr/local/bin/antsibull-changelog lint
-/usr/local/bin/ansible-lint roles/* -x 204
+ansible-test sanity --skip-test import
```
mwperina, line 73: Please add a comment why we need to skip import test

automation/build.sh
```diff
@@ -70,9 +70,9 @@ cd $COLLECTION_DIR
...
+antsibull-changelog lint
+ansible-lint roles/* -x 204
```
mwperina, line 75: Please add a comment why we need to skip this 204 test

tests/sanity/ignore-2.12.txt
```diff
@@ -0,0 +1,13 @@
+automation/build.sh shebang!skip
+build.sh shebang!skip
+changelogs/fragments/.gitignore changelog!skip
```
mwperina, line 3: Why do we need .gitignore under changelogs/fragments? Wouldn't it be better to remove it?
https://github.com/oVirt/ovirt-ansible-collection/pull/398

mwperina (changes requested):

mwperina (commented):
+1

mwperina (approved):

PR #409, opened 2021-12-15
roles: hosted_engine_setup: Adjust files permissions

arachmani:
Adjusting files permissions for supporting umask 077.

For security purposes we decided to use 770/660 permissions for the localvm directory, this directory contains the appliance image, and for creating the engine VM we must have its user/group configured for `qemu`, otherwise `qemu` user won't have search permissions for the appliance image (see also https://libvirt.org/drvqemu.html#posix-users-groups).
Therefore we have decided to use `qemu` group instead of `kvm`.

Bug-Url: https://bugzilla.redhat.com/2020620

### DIFF
--- a/roles/hosted_engine_setup/tasks/bootstrap_local_vm/02_create_local_vm.yml
+++ b/roles/hosted_engine_setup/tasks/bootstrap_local_vm/02_create_local_vm.yml
@@ -23,13 +23,6 @@
       - name: Set local vm dir path
         set_fact:
           he_local_vm_dir: "{{ otopi_localvm_dir.path }}"
-      - name: Fix local VM directory permission
-        file:
-          state: directory
-          path: "{{ he_local_vm_dir }}"
-          owner: vdsm
-          group: kvm
-          mode: 0775
       - include_tasks: ../install_appliance.yml
         when: he_appliance_ova is none or he_appliance_ova|length == 0
       - name: Register appliance PATH
@@ -82,6 +75,7 @@
         template:
           src: "{{ item.src }}"
           dest: "{{ item.dest }}"
+          mode: 0640
         with_items:
           - {src: templates/user-data.j2, dest: "{{ he_local_vm_dir }}/user-data"}
           - {src: templates/meta-data.j2, dest: "{{ he_local_vm_dir }}/meta-data"}
@@ -93,6 +87,14 @@
           {{ he_local_vm_dir }}/network-config
         environment: "{{ he_cmd_lang }}"
         changed_when: true
+      - name: Fix local VM directory permission
+        file:
+          state: directory
+          path: "{{ he_local_vm_dir }}"
+          owner: vdsm
+          group: qemu
+          recurse: yes
+          mode: u=rwX,g=rwX,o=
       - name: Create local VM
         command: >-
           virt-install -n {{ he_vm_name }}Local --os-variant rhel8.0 --virt-type kvm --memory {{ he_mem_size_MB }}
--- a/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
+++ b/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
@@ -155,8 +155,6 @@
       of="{{ he_conf_disk_path }}"
     environment: "{{ he_cmd_lang }}"
     become: true
-    become_user: vdsm
-    become_method: sudo
     changed_when: true
     args:
       warn: false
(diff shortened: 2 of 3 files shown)


mwperina (approved):
Looks good, although I wonder if we shouldn't try to use 770 or 660 masks

mwperina:
@didib

didib:
I personally think that 0770/0660 are mandatory, and if they do not work, we should find out why and fix that. The image includes sensitive stuff (hashed root password) and might include other sensitive stuff from user customization.

roles/hosted_engine_setup/tasks/bootstrap_local_vm/02_create_local_vm.yml
```diff
@@ -63,6 +63,12 @@
           dest: "{{ he_local_vm_dir }}"
           extra_opts: ['--sparse']
       - include_tasks: ../get_local_vm_disk_path.yml
+      - name: Adjust local VM directory permissions
+        file:
+          state: directory
+          path: "{{ he_local_vm_dir }}"
+          recurse: yes
+          mode: 0775
```
didib, line 71: Why? Also in line 32 above.

roles/hosted_engine_setup/tasks/bootstrap_local_vm/02_create_local_vm.yml
```diff
@@ -82,6 +88,7 @@
         template:
           src: "{{ item.src }}"
           dest: "{{ item.dest }}"
+          mode: 0644
```
didib, line 91: same

roles/hosted_engine_setup/tasks/bootstrap_local_vm/02_create_local_vm.yml
```diff
@@ -93,6 +100,10 @@
           {{ he_local_vm_dir }}/network-config
         environment: "{{ he_cmd_lang }}"
         changed_when: true
+      - name: Adjust ISO file permissions
+        file:
+          path: "{{ he_local_vm_dir }}/seed.iso"
+          mode: 0644
```
didib, line 106: same

roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
```diff
@@ -135,6 +135,10 @@
       warn: false
     changed_when: true
     tags: ['skip_ansible_lint']
+  - name: Adjust configuration archive file permissions
+    file:
+      path: "{{ he_local_vm_dir }}/{{ he_conf_disk_details.disk.image_id }}"
+      mode: 0644
```
didib, line 141: same

didib (changes requested):

roles/hosted_engine_setup/tasks/bootstrap_local_vm/02_create_local_vm.yml
arachmani: Using 2770/0770/0660 fails with the following error:
`You will need to grant the 'qemu' user search permissions for the following directories: ['/var/tmp/localvm_ztub84q']`.
Therefore keeping 0755 for sub-directories, 0644 for files in the `localvm` directory, and moving this task right before `Create local VM` task.

roles/hosted_engine_setup/tasks/bootstrap_local_vm/02_create_local_vm.yml
arachmani: Changed to 0640.

roles/hosted_engine_setup/tasks/bootstrap_local_vm/02_create_local_vm.yml
arachmani: This task has been removed, but still keep it 0644 for the iso file (as part of `Fix local VM directory permission` task) otherwise the deployment fails with the same error.

roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
arachmani: same

roles/hosted_engine_setup/tasks/bootstrap_local_vm/02_create_local_vm.yml
arachmani: Set to 0770/0660 and qemu group instead of kvm

arachmani:
Verified and opened https://bugzilla.redhat.com/show_bug.cgi?id=2035631

didib (approved):
Code looks fine to me. Perhaps worth documenting our considerations in the commit message or in comments.

PR #411, opened 2021-12-29
roles: hosted_engine_setup: Add an option to define OpenSCAP security profile name

arachmani:
Currently, we support only stig security profile.
This PR allows defining the security profile name,
available profile names: stig(default), pci-dss

Bug-Url: https://bugzilla.redhat.com/2029830


roles/hosted_engine_setup/tasks/apply_openscap_profile.yml
```diff
@@ -10,10 +10,10 @@
   stat:
     path: "{{ oscap_datastream }}"
   register: oscap_ds_stat
-- name: Set default OpenSCAP profile
+- name: Set OpenSCAP profile
   shell: >-
     set -euo pipefail && oscap info --profiles {{ oscap_datastream }} |
-    grep -Ei "(standard|disa)" | sort | tail -1 | cut -d':' -f1
+    grep -Ei "({{ he_openscap_profile_name }}:)" | sort | tail -1 | cut -d':' -f1
```
almusil, line 16: Why user cannot simply specify the whole name instead of this filtering?
Also this won't work for disa because there are two disa profiles: 
- xccdf_org.ssgproject.content_profile_stig
- xccdf_org.ssgproject.content_profile_stig_gui

And the command will actually select the gui one, which is not what we want.

arachmani: > Why user cannot simply specify the whole name instead of this filtering?

The user simply needs to specify `stig` or `pci-dss` which IMO is easier than specifying the entire name.
Why do we need this prefix if it's the same for all profiles?

> Also this won't work for disa because there are two disa profiles:
> 
> * xccdf_org.ssgproject.content_profile_stig
> * xccdf_org.ssgproject.content_profile_stig_gui
> 
> And the command will actually select the gui one, which is not what we want.

This works fine, please be aware of the `:` after the `{{ he_openscap_profile_name }}`.

almusil: > > Why user cannot simply specify the whole name instead of this filtering?
> 
> The user simply needs to specify `stig` or `pci-dss` which IMO is easier than specifying the entire name. Why do we need this prefix if it's the same for all profiles?
>
> > Also this won't work for disa because there are two disa profiles:
> > 
> > * xccdf_org.ssgproject.content_profile_stig
> > * xccdf_org.ssgproject.content_profile_stig_gui
> > 
> > And the command will actually select the gui one, which is not what we want.
> 
> This works fine, please be aware of the `:` after the `{{ he_openscap_profile_name }}`.

Ok did not notice that, makes sense then, thanks

mwperina: OK, let's hope we won't need any other profile in the near future.

roles/hosted_engine_setup/README.md
```diff
@@ -43,6 +43,7 @@ Ansible version >= 2.9.21 and < 2.10.0
 | he_force_ip4 | false | Force resolving engine FQDN to ipv4 only using DNS server |
 | he_force_ip6 | false | Force resolving engine FQDN to ipv6 only using DNS server |
 | he_apply_openscap_profile | false | apply a default OpenSCAP security profile on HE VM |
+| he_openscap_profile_name | stig | OpenSCAP profile name, available options: *stig*, *pci-dss*. Requires `he_apply_openscap_profile` to be `True` |
```
mwperina, line 46: Hmm, do we want really stig to be the default profile?

@michalskrivanek

arachmani: Currently, users just specify the `he_apply_openscap_profile` bool parameter, and if `true` the default is stig.
So we can keep this behavior by setting `stig` as default (it's only when `he_apply_openscap_profile` is `True` and `he_openscap_profile_name` not specified).

mwperina: OK, makes sense

mwperina (approved):
+1

roles/hosted_engine_setup/tasks/apply_openscap_profile.yml
```diff
@@ -22,8 +22,8 @@
   command: >-
     oscap xccdf eval --profile {{ oscap_profile.stdout }} --remediate
     --report /root/openscap-report.html {{ oscap_datastream }}
-  failed_when: false
-  ignore_errors: true
+  register: oscap_applied
+  failed_when: oscap_applied.rc == 1
```
almusil, line 26: Actually we should differentiate between rc=1 and rc=2. 
According to oscap man page:
`oscap returns 0 if all rules pass. If there is an error during evaluation, the return code is 1. If there is at least one rule with either fail or unknown result, oscap-scan finishes with return code 2.`

arachmani: Indeed, AFAICT when applying security profile on HE vm (and also on RHVH via anconda) there are many rules that fail, this is why deployment will fail only when `rc == 1`.
I also opened https://bugzilla.redhat.com/show_bug.cgi?id=1983476.

didib (approved):

roles/hosted_engine_setup/tasks/apply_openscap_profile.yml
arachmani: I added a comment

arachmani:
Ci fails on https://github.com/ansible-community/ansible-lint/issues/1795

mwperina:
Workaround for ansible-lint posted: https://github.com/oVirt/ovirt-ansible-collection/pull/413

PR #412, opened 2022-01-05
workflows: Create DNF repo metadata for artifacts

tinez:
In accord to GitHub migration documentation [1] we should generate
DNF repository metadata for the built RPMs, so the artifact archive
can be used with OST.

[1] https://www.ovirt.org/develop/developer-guide/migrating_to_github.html

Signed-off-by: Marcin Sobczyk <msobczyk@redhat.com>


mwperina (approved):
+1

PR #415, opened 2022-01-12
roles: cluster_upgrade: Add progress tracking/reporting

sjd78:
Since cluster upgrade is a very long running process, adding progress tracking allows users and admins to easily track status.  Progress tracking is done by posting well formatted events.

Note: `roles/cluster_upgrade/tasks/upgrade.yml` was converted to a block.  This allows calculating var the tasks for a host upgrade needs without using `set_fact` style global facts.  The only annoyance is having to put them at the bottom of the file below all of the tasks.

Bug-Url: https://bugzilla.redhat.com/2040474
Signed-off-by: Scott J Dickerson <sdickers@redhat.com>

### DIFF
--- a/roles/cluster_upgrade/tasks/main.yml
+++ b/roles/cluster_upgrade/tasks/main.yml
@@ -231,15 +231,6 @@
             progress: 95
             description: "host upgrades are done (successful or not), restarting non-migratable VMs"
 
-        # TODO: These VMs aren't explicity stopped anywhere...should they be?
-        - name: Start again stopped VMs
-          ovirt_vm:
-            auth: "{{ ovirt_auth }}"
-            name: "{{ item }}"
-            state: running
-          ignore_errors: "yes"
-          loop: "{{ stopped_vms | default([]) | flatten(levels=1) }}"
-
         - name: Start again pin to host VMs
           ovirt_vm:
             auth: "{{ ovirt_auth }}"
--- a/roles/cluster_upgrade/tasks/main.yml
+++ b/roles/cluster_upgrade/tasks/main.yml
@@ -24,122 +24,177 @@
       tags:
         - always
 
+    - name: progress 0% - need to do info lookups
+      include_tasks: log_progress.yml
+      vars:
+        progress: 0
+        description: "gathering cluster info"
+
     - name: Get API info
       ovirt_api_info:
         auth: "{{ ovirt_auth }}"
-      register: api_info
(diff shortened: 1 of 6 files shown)


roles/cluster_upgrade/tasks/main.yml
```diff
@@ -164,47 +225,59 @@
             scheduling_policy_properties: "{{ cluster_scheduling_policy_properties }}"
           when: use_maintenance_policy and cluster_policy.changed | default(false)
 
+        - name: progress 95% - host upgrades are done (successful or not), only need to start VMs that were stopped by the playbook
+          include_tasks: log_progress.yml
+          vars:
+            progress: 95
+            description: "host upgrades are done (successful or not), restarting non-migratable VMs"
+
+        # TODO: These VMs aren't explicity stopped anywhere...should they be?
```
mnecas, line 234: If I remember correctly the VMs are stopped when doing the host upgrade by themselves so I think we should remove the parameter `stopped_vms` it has not been in use in this role for multiple years. We handle here only non-migratable VMs

mnecas (changes requested):

roles/cluster_upgrade/tasks/main.yml
```diff
@@ -24,137 +24,198 @@
...
+      no_log: false
       uri:
-        url: "{{ ovirt_auth.url }}/clusters/{{ cluster_info.ovirt_clusters[0].id }}/upgrade"
+        url: "{{ ovirt_auth.url }}/clusters/{{ cluster_id }}/upgrade"
         method: POST
         body_format: json
         validate_certs: false
         headers:
           Authorization: "Bearer {{ ovirt_auth.token }}"
+          Correlation-Id: "{{ engine_correlation_id | default(omit) }}"
```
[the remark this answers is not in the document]
mnecas: my mistake already there

roles/cluster_upgrade/tasks/main.yml
sjd78: Since even the cluster upgrade wizard does not use that option, I will just remove it.

mnecas: After discussion, we have decided to keep it for different PR. Either to reimplement it or remove it. I have added it to my todo list.
Updated PR and will merge after the CI.

mnecas (approved):
LGTM

PR #425, opened 2022-02-01
cluster_upgrade: shutdown vms only on pinned hosts

mnecas:
Issue: https://bugzilla.redhat.com/show_bug.cgi?id=2049286
Running cluster-upgrade updated all hosts instead of the selected.


mwperina (changes requested):
Please add commit message describing tthe fix and also mention relevant bug https://bugzilla.redhat.com/show_bug.cgi?id=2049286

Other than that it looks good

mnecas:
Verified

mwperina (approved):
+1

PR #426, opened 2022-02-07
roles: hosted_engine_setup: Support disa stig profile

arachmani:
DISA STIG profile uses fapolicyd that blocks ansible command execution as non-root
(see: https://bugzilla.redhat.com/show_bug.cgi?id=1903549)
Adding a rule to /etc/fapolicyd/rules.d to allow ansible command execution as `vdsm` user.
The rule will be removed once it's no longer required.

Bug-Url: https://bugzilla.redhat.com/2020620

### DIFF
--- a/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
+++ b/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
@@ -149,25 +149,29 @@
     with_items:
       - {src: "{{ he_local_vm_dir }}/vm.conf", dest: /var/run/ovirt-hosted-engine-ha}
       - {src: "{{ he_local_vm_dir }}/hosted-engine.conf", dest: /etc/ovirt-hosted-engine/}
-  - name: Verify fapolicyd.rules file
-    stat:
-      path: /etc/fapolicyd/fapolicyd.rules
-    register: fapolicy_rules
-  - name: Add rule to fapolicyd.rules file
-    blockinfile:
-      path: /etc/fapolicyd/fapolicyd.rules
-      block: |
-        # temporary entry added by hosted-engine-setup for running commands as vdsm user
-        allow perm=any uid=36 : dir=/var/tmp/
-      insertbefore: "^# Allow any program to open trusted language files"
-      marker: "## {mark} HOSTED-ENGINE-SETUP MANAGED BLOCK ##"
-      backup: yes
-    when: fapolicy_rules.stat.exists
-  - name: Restart fapolicyd service
-    service:
+  - name: Check fapolicyd status
+    systemd:
       name: fapolicyd
-      state: restarted
-    when: fapolicy_rules.stat.exists
+    register: fapolicyd_s
+  - name: Set fapolicyd rules path
+    set_fact:
+      fapolicyd_rules_dir: /etc/fapolicyd/rules.d
+  - name: Verify fapolicyd/rules.d directory
+    stat:
+      path: "{{ fapolicyd_rules_dir }}"
+    register: fapolicy_rules
+  - name: Add rule to fapolicy
+    block:
+      - name: Add rule to /etc/fapolicyd/rules.d
+        copy:
+          src: 35-allow-ansible-for-vdsm.rules
+          dest: "{{ fapolicyd_rules_dir }}"
+          mode: 0644
+      - name: Restart fapolicyd service
+        service:
+          name: fapolicyd
+          state: restarted
+    when: fapolicyd_s.status.SubState == 'running' and fapolicy_rules.stat.exists
   - name: Copy configuration archive to storage
     command: >-
       dd bs=20480 count=1 oflag=direct if="{{ he_local_vm_dir }}/{{ he_conf_disk_details.disk.image_id }}"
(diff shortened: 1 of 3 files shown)


roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
```diff
@@ -162,8 +162,6 @@
     command: dd bs=1M count=1024 oflag=direct if=/dev/zero of="{{ he_metadata_disk_path }}"
     environment: "{{ he_cmd_lang }}"
     become: true
-    become_user: vdsm
-    become_method: sudo
```
didib, line 166: Who will be the owner of the file, then? Please explain in a code comment or the commit message why we needed to be vdsm and now we do not. Same for below.

nirs: Why are you zeroing 1g in this volume?

If you want to allocate space, use:

   fallocate --length ... /path/to/file

If you want to clean up a block device, use:

    blkdiscard --zeroout /path/to/device
    blkdiscard /path/to/device

roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
```diff
@@ -149,19 +149,42 @@
...
+  - name: Verify fapolicyd.rules file
+    stat:
+      path: /etc/fapolicyd/fapolicyd.rules
+    register: fapolicy_rules
+  - name: Add rule to fapolicyd.rules file
+    blockinfile:
+      path: /etc/fapolicyd/fapolicyd.rules
+      block: |
+        # temporary entry added by hosted-engine-setup for running commands as vdsm user
+        allow perm=any uid=36 : dir=/var/tmp/
```
mwperina, line 161: Instead of doing this hack, can't we just use different directory where this is allowed?

roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
```diff
@@ -149,12 +149,33 @@
...
+      insertbefore: "^# Allow any program to open trusted language files"
+      marker: "## {mark} HOSTED-ENGINE-SETUP MANAGED BLOCK ##"
+      backup: yes
+    when: fapolicy_rules.stat.exists
+  - name: Restart fapolicyd service
+    service:
+      name: fapolicyd
+      state: restarted
+    when: fapolicy_rules.stat.exists
```
almusil, line 170: This checks if the config exists, but the service might not be running it might be just installed.

roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
arachmani: > Why are you zeroing 1g in this volume?
> 
> If you want to allocate space, use:
> 
> fallocate --length ... /path/to/file
> 
> If you want to clean up a block device, use:
> 
> ```
> blkdiscard --zeroout /path/to/device
> blkdiscard /path/to/device
> ```

It's not really related to this PR and I don't know the historical reason behind that, I'll check that, thanks.

roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
arachmani: Ansible files located in /var/tmp, not sure if we can and if it's a good idea to change it.
fapolicy added something similar for root (see https://bugzilla.redhat.com/show_bug.cgi?id=1746464).

roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
arachmani: Thanks, I'll also check if the service is running.

roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
arso: Please take a look into new feature in fapolicy [rules.d ](https://github.com/linux-application-whitelisting/fapolicyd/issues/117) that is going to be introduced in upcoming release fapolicy-1.1  (in RHEL 8.6).
 It's not there yet but I got information from its core developer that the feature will land in RHEL 8.6 nightly builds today/tomorrow.

An example (not yet verified) of the usage is in the PR I have prepared for Ovirt Engine's Wilddfly runtime tmp directory [here](https://github.com/linux-application-whitelisting/fapolicyd/issues/117)

arachmani: fapolicyd blocks ansible commands (located in /var/tmp) execution as non-root user.
For more details see https://bugzilla.redhat.com/1903549

didib (commented):
If urgent, and current patch works, ok for me to merge. That said, I'd like to see explanations, in code comments or at least git commit messages, for:

1. Why can't we run this copy/dd as root and change permissions/context/whatever later as needed, to make virt-install happy
2. Why can't we make ansible write to some other location - perhaps using ansible_remote_tmp - which is already allowed for vdsm.

If one of these was not tried yet, I'd personally prefer it, if it works, from a technical POV (ignoring urgency/priority/capacity/etc.).

arso (approved):
lgtm - however, I mostly focused on checking fapolicy configuration and the necessity to restart fapolicyd service if it was running before making updates.

mwperina (approved):
+1

arachmani:
> If urgent, and current patch works, ok for me to merge. That said, I'd like to see explanations, in code comments or at least git commit messages, for:
> 
> 1. Why can't we run this copy/dd as root and change permissions/context/whatever later as needed, to make virt-install happy
> 2. Why can't we make ansible write to some other location - perhaps using ansible_remote_tmp - which is already allowed for vdsm.
> 
> If one of these was not tried yet, I'd personally prefer it, if it works, from a technical POV (ignoring urgency/priority/capacity/etc.).

1. virt-install is already happy thanks to https://github.com/oVirt/ovirt-ansible-collection/pull/409.
   we need to make sure we run the commands with the right owner and not change the owner by mistake 
   I'm not a storage guy and not sure about all the options we have, but AFAIU to use root user on NFS with rootsquash could be problematic.
2. I didn't check that and am not sure if it is indeed a better solution.

PR #430, opened 2022-02-23
repositories: fix force flag on subscription-manager

mnecas:
Issue: The force flag on subscription-manager was always present. It was just checking if the  `ovirt_repositories_force_register` was defined which was in default so it was always present.


mwperina (approved):
+1

PR #433, opened 2022-02-26
Add delay after patching the current host

dhoogfr:
Add delay after patching the current host, before continuing with the next host

After patching a host, it takes some time before that host (after rebooting) is a valid target again for guests to be migrated to.
With a 2 node cluster, this can cause the patching of the second host to fail because the guests cannot be migrated of that host or to cause the guests to be shutdown (depending on the used parameters).
This can be avoided by adding a delay (controlled by the wait_before_next_upgrade parameter, defaulting to 5 minutes) before continuing with the next host


dhoogfr:
Hi,

I noticed that one build failed because of a missing changelog?
Do I need to provide something for this?

dupondje:
No more updates on this since long time.
Closing. Feel free to re-open if you want to continue working on it!

PR #435, opened 2022-02-28
[WIP] roles: hosted_engine_setup: Replace community modules

arachmani:
We need to replace community modules since they are not included in
ansible >= 2.11

Bug-Url: https://bugzilla.redhat.com/2022620

### DIFF
--- a/roles/hosted_engine_setup/tasks/create_storage_domain.yml
+++ b/roles/hosted_engine_setup/tasks/create_storage_domain.yml
@@ -41,11 +41,15 @@
       auth: "{{ ovirt_auth }}"
     register: datacenter_facts
   - name: Fetch Datacenter ID
-    set_fact: >-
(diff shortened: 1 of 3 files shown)


roles/hosted_engine_setup/tasks/create_storage_domain.yml
```diff
@@ -41,11 +41,15 @@
       auth: "{{ ovirt_auth }}"
     register: datacenter_facts
   - name: Fetch Datacenter ID
-    set_fact: >-
-      datacenter_id={{ cluster_facts.ovirt_clusters|json_query("[?id=='" + cluster_id + "'].data_center.id")|first }}
```
mnecas, line 45: Watch out this gets you the first item in the list your solution gives you the last.
(not sure if it matters here, just note)

arachmani:
After discussing with @mnecas we decided to copy the json_query filter to the ovirt.ovirt collection.
Please see: https://github.com/oVirt/ovirt-ansible-collection/pull/436

PR #443, opened 2022-03-07
roles: hosted_engine_setup: Use cat command instead of lookup

arachmani:
We get "could not locate file in lookup:" error, probably because
lookups execute and are evaluated on the Ansible control machine
and not on the remote machine


jirimacku:
Fix tested with downstream automation - passed.

sandrobonazzola (approved):

PR #446, opened 2022-03-16
Add editorconfig

mnecas:
The .editorconfig is used for configuring the IDE/code editor, which needs to have support. We can specify for example tabulator sizes for specific files. To see more information you can checkout https://editorconfig.org/


mwperina (approved):
+1

PR #447, opened 2022-03-16
Automation: Use oVirt buildcontainer

mnecas:
Use the oVrit build container from quay.io/ovirt/buildcontainer for the automation


michalskrivanek (changes requested):
- you don't need to install createrepo_c, rpm-build, etc.
- add the deps on line 20 as BuildRequires - they will be included in buildcontainer then
- use https://github.com/oVirt/upload-rpms-action for upload action
- consider matrix config/template from https://github.com/oVirt/buildcontainer/README.md

PR #448, opened 2022-03-18
Automation: Fix ansible build workflow

mnecas:
The ansible-core is provided by default repos from el8stream we don't need to use epel.


mwperina (approved):
+1

PR #452, opened 2022-03-23
roles: hosted_engine_setup: Allow postgres user in fapolicyd

arachmani:
DISA STIG profile uses fapolicyd that blocks ansible command execution as non-root
(see: https://bugzilla.redhat.com/show_bug.cgi?id=1903549)
Adding a rule to /etc/fapolicyd/rules.d to allow ansible command execution as postgres user.
The rule will be removed once it's no longer required.

Bug-Url: https://bugzilla.redhat.com/2066811


mwperina (approved):
+1

mwperina:
Upon offline discussion we have decided to use https://github.com/oVirt/ovirt-ansible-collection/pull/453 instead

PR #453, opened 2022-03-23
Replace calls to psql as postgres with engine_psql.sh

didib:
engine_psql.sh is a script provided by the engine to access its DB.
It does not su/sudo to postgres but uses the configured credentials.
Do this for FIPS support, where psql as postgres fails.

### DIFF
--- a/roles/hosted_engine_setup/defaults/main.yml
+++ b/roles/hosted_engine_setup/defaults/main.yml
@@ -36,6 +36,8 @@ he_webui_forward_port: 6900  # by default already open for VM console
 he_reserved_memory_MB: 512
 he_avail_memory_grace_MB: 200
 
+engine_psql: /usr/share/ovirt-engine/dbscripts/engine-psql.sh
+
 he_host_ip: null
 he_host_name: null
 he_host_address: null
--- a/changelogs/fragments/453-replace-calls-to-psql-as-postgres-with-engine_psql.yml
+++ b/changelogs/fragments/453-replace-calls-to-psql-as-postgres-with-engine_psql.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - hosted_engine_setup - Replace calls to psql as postgres with engine_psql.sh (https://github.com/oVirt/ovirt-ansible-collection/pull/453).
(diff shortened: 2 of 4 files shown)


roles/hosted_engine_setup/defaults/main.yml
```diff
@@ -36,6 +36,8 @@ he_webui_forward_port: 6900  # by default already open for VM console
 he_reserved_memory_MB: 512
 he_avail_memory_grace_MB: 200
 
+engine_psql: /usr/share/ovirt-engine/dbscripts/engine-psql.sh
```
arachmani, line 39: Using an external script might be tricky to debug in case of changes/failure.
IMO going with https://github.com/oVirt/ovirt-ansible-collection/pull/452 is a more conservative approach.

arachmani, line 39: Using an external script might be tricky to debug in case of changes/failure.
IMO going with https://github.com/oVirt/ovirt-ansible-collection/pull/452 is a more conservative approach.

michalskrivanek: it's far less changes. this is a slightly controversial script indeed, OTOH we carry it for ages and plenty people using it..I think it's elegant to do it this way and not require any changes for ansible/fapolicyd

arachmani: Basically, the code for adding a rule in fapolicyd already exists (for vdsm), I just make it generic and use it for postgres.

didib: IMO it's the other way around: If something breaks while we use engine-psql.sh, we can more quickly fix that. If something breaks while we change fapolicyd, this might require changes there, which will take more time. Also, I do not know fapolicyd, never read about it. I consider it mentally to be similar to selinux policy - should work out-of-the-box, and if something breaks due to it, should be fixed there. The "thing" that gets broken should IMO not have to bother with this. So if RHEL+stig now has a policy that somehow forbids su to postgres to access a random DB (a point which IMO is at least controversial, but it seems like it's "the policy"), and we have clear, straightforward means to not do that - but use the credentials we configured ourselves, we should pick latter. 'su - postgres' here IMO was always a hack - we should always have use something like engine-psql.

didib: I still prefer current, but no strong feelings.

mnecas: I would go with this change as well does not change the code that much.

michalskrivanek (approved):

mwperina (approved):
+1

PR #454, opened 2022-03-23
Repositories: fix dnf module variable

mnecas:
Use the proper variable in repositories to set dnf modules when using satellite


mwperina (approved):
+1

PR #459, opened 2022-03-28
roles: hosted_engine_setup: Fix call to engine-psql for vds_spm_id

didib:
A recent patch replaced 'psql -d engine' with 'engine-psql'. One call
also passed '-t' (tuples-only), but that patch dropped this option, thus
breaking parsing of the output. Restore this option.


mnecas (approved):
lgtm

PR #463, opened 2022-03-29
Add storage format

didib:
Allow passing storage_format as is allowed by the Python SDK.

Bug-Url: https://bugzilla.redhat.com/1932147

### DIFF
--- a/changelogs/fragments/463-hosted_engine_setup-add-and-set-storage-format.yml
+++ b/changelogs/fragments/463-hosted_engine_setup-add-and-set-storage-format.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - hosted_engine_setup - Add an option to set the storage format when createing a storage domain and use it (https://github.com/oVirt/ovirt-ansible-collection/pull/463).
--- a/plugins/modules/ovirt_storage_domain.py
+++ b/plugins/modules/ovirt_storage_domain.py
@@ -230,6 +230,10 @@ options:
             - "If I(True) storage domain blocks will be discarded upon deletion. Enabled by default."
             - "This parameter is relevant only for block based storage domains."
         type: bool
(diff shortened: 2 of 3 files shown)


sandrobonazzola (approved):
LGTM but I would prefer some more people to review this one

arachmani (commented):
The code LGTM, can you please add a changelog?

didib:
Verified in [1] with [2] and current.

The commit messages in [2] are not very up to date - to get to this success, I also had to put host-0 to maintenance and remove it, prior to taking the backup.

Without current, something like [1] failed with the error message in the bug. With it, HE deploy succeeded. [1] failed later on adding host-1 (because it already exists).

So current PR can be merged as-is. If you prefer to not include the middle patch, I don't mind removing it, or you can remove it too if you want.

[1] https://redir.apps.ovirt.org/dj/job/ds-ost-baremetal_manual/32505

[2] https://github.com/oVirt/ovirt-system-tests/pull/93

arachmani:
> So current PR can be merged as-is. If you prefer to not include the middle patch, I don't mind removing it, or you can remove it too if you want.

The middle commit contains code that looks similar to `create_storage_domain.yml`.
It's also not that trivial how to use it, I think we should remove it or perhaps combine them and add an explanation.

didib:
> The middle commit contains code that looks similar to `create_storage_domain.yml`. It's also not that trivial how to use it, I think we should remove it or perhaps combine them and add an explanation.

I'd remove it.

didib:
Ran OST, it passed HE deploy and failed later like before, as expected. So looks good to me, can be merged.

https://redir.apps.ovirt.org/dj/job/ds-ost-baremetal_manual/32907

arachmani (approved):
LGTM

PR #466, opened 2022-03-31
fix creating a RAW vm from a COW template

olirdx:
On a FC storage domain, creating a RAW vm (disk_format: "raw") from a COW template with vm_infra role gives as result : "Cannot add VM. Disk configuration (RAW Sparse backup-None) is incompatible with the storage domain type." 
I could be wrong on diagnostic but this seems to be caused by the unset "sparse" option. This commit fix it on our infrastructure.


mnecas (approved):
Thanks so much for the patch! LGTM

PR #475, opened 2022-04-06
roles: hosted_engine_setup: Use correct admin user with keycloak

didib:


michalskrivanek:
i think i can test that in ost.....

mwperina:
This is a blocker for oVirt 4.5.0 GA. Could you please verify, so we can merge and perform a new ovirt-ansible-collection release on April 19th to get into oVirt 4.5.0 GA compose?

sandrobonazzola (approved):
LGTM

didib:
@michalskrivanek I pushed a similar patch to ovirt-system-tests which you reverted. Why? If it wasn't good enough there, perhaps it's also not good enough here.

didib:
https://redir.apps.ovirt.org/dj/job/ds-ost-baremetal_manual/33877/

michalskrivanek:
> @michalskrivanek I pushed a similar patch to ovirt-system-tests which you reverted. Why? If it wasn't good enough there, perhaps it's also not good enough here.

because we don't have OVESETUP_KEYCLOAK_CORE/enable consistently set up in /etc/ovirt-engine-setup.conf, and we don't have it at all before setup is finished. 

in HE at 04_engine_final_tasks.yml we can query for the resulting value. But in order to use keycloak we'd need to revert the disablement in https://github.com/oVirt/ovirt-appliance/pull/54/files

didib:
#487 is an alternative to current, perhaps more reliable

didib:
Won't work either, because we need to know if keycloak is enabled also for later runs, not just bootstrap_local_vm

didib:
> > @michalskrivanek I pushed a similar patch to ovirt-system-tests which you reverted. Why? If it wasn't good enough there, perhaps it's also not good enough here.
> 
> because we don't have OVESETUP_KEYCLOAK_CORE/enable consistently set up in /etc/ovirt-engine-setup.conf,

That's why I added 'ignore_errors: true' and check rc. Isn't this enough?

>  and we don't have it at all before setup is finished.

Indeed, which might be an issue for OST, but not for current patch. For OST, IMO we should:
1. Have it (keycloak_enabled) set by the env (e.g. user running OST) and not by the code (e.g. guessing)
2. Make "everything" adhere to the decision in (1.), including basic suite (running engine-setup directly) and he-basic (ideally perhaps via a patch to ovirt-ansible-collection. Again, not by guessing).

> 
> in HE at 04_engine_final_tasks.yml we can query for the resulting value. But in order to use keycloak we'd need to revert the disablement in https://github.com/oVirt/ovirt-appliance/pull/54/files

Of course. That was just a temporary quick and simple hack.

didib:
Closing current in favor of #488. They use similar logic where they match, but 488 does more, and we are now not in a hurry, as oVirt 4.5.0 was already released and keycloak integration was postponed to 4.5.1.

PR #480, opened 2022-04-07
Obsolete ansible 2.9

mnecas:
Since ansible-core 2.12.3 the package does not obsolete but conflicts with ansible<2.10 and because of this, the upgrade fails.
The solution for this is to add our obsolete. When doing an upgrade from now it will remove the ansible 2.9 and install ansible-core 2.12.


michalskrivanek (approved):

michalskrivanek:
merging to get automation feedback when using COPR

PR #486, opened 2022-04-14
ovirt_vm: Fix virtio_scsi_enabled argument

dupondje:
The virtio_scsi_enabled arg has a typo, causing the following error:
  File \"/tmp/ansible_ovirt_vm_payload_8un8pxki/ansible_ovirt_vm_payload.zip/ansible_collections/ovirt/ovirt/plugins/modules/ovirt_vm.py\", line 2671, in main
  File \"/tmp/ansible_ovirt_vm_payload_8un8pxki/ansible_ovirt_vm_payload.zip/ansible_collections/ovirt/ovirt/plugins/module_utils/ovirt.py\", line 644, in create
    if not self.update_check(entity):
  File \"/tmp/ansible_ovirt_vm_payload_8un8pxki/ansible_ovirt_vm_payload.zip/ansible_collections/ovirt/ovirt/plugins/modules/ovirt_vm.py\", line 1689, in update_check
  File \"/tmp/ansible_ovirt_vm_payload_8un8pxki/ansible_ovirt_vm_payload.zip/ansible_collections/ovirt/ovirt/plugins/modules/ovirt_vm.py\", line 1787, in _update_check
AttributeError: 'NoneType' object has no attribute 'enabled'


plugins/modules/ovirt_vm.py
```diff
@@ -1784,7 +1784,7 @@ def check_custom_compatibility_version():
             equal(self.param('serial_policy'), str(getattr(entity.serial_number, 'policy', None))) and
             equal(self.param('serial_policy_value'), getattr(entity.serial_number, 'value', None)) and
             equal(self.param('numa_tune_mode'), str(entity.numa_tune_mode)) and
-            equal(self.param('virtio_scsi_enabled'), entity.virtio_scsi.enabled) and
+            equal(self.param('virtio_scsi_enabled'), getattr(entity.virtio_scsi, 'enabled', None)) and
             equal(self.param('multi_queues_enabled'), entity.multi_queues_enabled) and
```
ahadas, line 1788: don't we need to do something similar here as well?

dupondje: I don't think so, as multi_queues_enabled for example always exists in the entity.
For virtio_scsi we need this, because the object is not always created, and thus enabled is checked of None.

See https://github.com/oVirt/ovirt-ansible-collection/blob/dfb33536f7dcc73870aa6db9657775522669b67b/plugins/modules/ovirt_vm.py#L1585

dupondje: But we might need it in some other places:

https://github.com/oVirt/ovirt-ansible-collection/blob/dfb33536f7dcc73870aa6db9657775522669b67b/plugins/modules/ovirt_vm.py#L1765 Is one for example

But also: https://github.com/oVirt/ovirt-ansible-collection/blob/dfb33536f7dcc73870aa6db9657775522669b67b/plugins/modules/ovirt_vm.py#L1771

So yes, needs some more improvements

dupondje: Fixed in new push

mnecas (approved):
LGTM

PR #488, opened 2022-04-14
Configure keycloak

didib:


roles/hosted_engine_setup/tasks/bootstrap_local_vm/04_engine_final_tasks.yml
```diff
@@ -48,6 +48,20 @@
...
+    ignore_errors: true
+  - name: Set admin username
+    set_fact:
+      he_admin_username: >-
+        {{ 'admin@ovirt@internalsso'
+            if keycloak_configured_out.rc == 0 and keycloak_configured_out.stdout_lines[0] == 'True'
+            else 'admin@internal'
+        }}
+  - name: Pass engine admin user to otopi
+    set_fact:
```
didib, line 63: This does not work, because ohe-setup's 1_otopi_json.py looks only at 'register' results. Not sure I can 'register:' the results of 'set_fact' or what this even means. Now searched a bit and could not find anything helpful. Need to do something - either hack (e.g. run command: 'echo $stuff' and register: the results) or check if I can register: set_fact or something similar.

arachmani: Maybe register the output of the `Set admin username` task to `otopi_he_admin_username`.

didib: Done. Still didn't check previous, so not sure if current will work either. Let's see.

roles/hosted_engine_setup/tasks/bootstrap_local_vm/03_engine_initial_tasks.yml
```diff
@@ -53,6 +53,13 @@
     with_items:
       - "OVESETUP_DWH_DB/password=str:{{ he_dwh_db_password }}"
     when: he_dwh_db_password is defined
+  - name: Add keycloak line to answerfile
+    lineinfile:
+      path: /root/ovirt-engine-answers
+      line: "{{ item }}"
+    with_items:
+      - "OVESETUP_KEYCLOAK_CORE/enable=bool:{{ he_enable_keycloak }}"
+    when: he_enable_keycloak is defined
```
arachmani, line 62: `he_enable_keycloak` is always defined since you added it to `defaults/main.yml`

roles/hosted_engine_setup/tasks/bootstrap_local_vm/04_engine_final_tasks.yml
```diff
@@ -48,6 +48,18 @@
...
       - cloud-init
+  - name: Check if keycloak is configured
+    command: otopi-config-query query -k OVESETUP_KEYCLOAK_CORE/enable -f /etc/ovirt-engine-setup.conf
+    register: keycloak_configured_out
+    ignore_errors: true
+  - name: Set admin username
+    set_fact:
+      he_admin_username: >-
+        {{ 'admin@ovirt@internalsso'
+            if keycloak_configured_out.rc == 0 and keycloak_configured_out.stdout_lines[0] == 'True'
```
arachmani, line 59: AFAICT, we basically update otopi based on the ansible role and override the ansible role variables based on otopi.
Maybe update `he_admin_username` if `he_enable_keycloak` is true, and in otopi, update `he_enable_keycloak` to be false/true based on customer needs.

roles/hosted_engine_setup/README.md
```diff
@@ -32,6 +32,7 @@ Ansible version >= 2.9.21 and < 2.10.0
 | he_ansible_host_name | localhost | hostname in use on the first HE host (not necessarily the Ansible controller one) |
 | he_restore_from_file | null | a backup file created with engine-backup to be restored on the fly |
 | he_pki_renew_on_restore | false | Renew engine PKI on restore if needed |
+| he_enable_keycloak | true | Configure keycloak on the engine if possible |
```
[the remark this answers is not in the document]
michalskrivanek: that was the intention, it obsoletes the internal AAA

didib:
Relevant flows:

1. With [1], and with the appliance or the engine patched to include ovirt-engine-keycloak:
1.1 ovirt-hosted-engine-setup asks whether to configure keycloak (default to True), and passes the answer to current code as he_enable_keycloak var
1.2 Current code changes the engine-setup answer file to enable keycloak, then runs engine-setup
1.3 Current code, still in bootstrap_local_vm, checks if keycloak was enabled by engine-setup, and sets he_admin_username accordingly, also passing the result on to ovirt-hosted-engine-setup as otopi_he_admin_username. Please note that this test is not 100% proof - if ovirt-engine-keycloak (the package) is included (with postinstallfile=True for OVESETUP_KEYCLOAK_CORE/enable, which is), the result will be True, regardless of what the code there does.
1.4 ovirt-hosted-engine-setup passes he_admin_username as was provided to it, to the next ansible runs (create_storage_domain, create_target_vm, etc.) and so auth_sso should work well.

2. Without [1], I think that HE deploy will fail on the "next" (not first) ansible runs, due to missing he_admin_username. I don't mind giving it a default value, but perhaps it's better to fail.

3. With [1] but without ovirt-engine-keycloak on the appliance, I think, not sure, that everything will work, using the "old" admin username (admin@internal).

I only tested (1.) above, with [2]. [2] adds the package and Provides a 'Yes' answer to the question added in [1].

[1] https://github.com/oVirt/ovirt-hosted-engine-setup/pull/44
[2] https://github.com/oVirt/ovirt-system-tests/pull/133

roles/hosted_engine_setup/tasks/bootstrap_local_vm/04_engine_final_tasks.yml
mwperina: Well, doesn't it cover also the scenario when you pause HE installation, SSH to engine VM and disable keycloak? If so, then that's a good fail safe.

roles/hosted_engine_setup/tasks/bootstrap_local_vm/03_engine_initial_tasks.yml
mwperina: Removed

mwperina (approved):
LGTM

arso (approved):
lgtm

roles/hosted_engine_setup/tasks/bootstrap_local_vm/04_engine_final_tasks.yml
michalskrivanek: true. otherwise it's more or less equivalent, so let's keep it like this when already tested...

PR #489, opened 2022-04-24
roles: hosted_engine_setup: Allocate 128MiB instead of 1GiB for he_metadata

arachmani:
Each host in hosted-engine cluster is going to write to metadata volume 
in his own reserved area (offset).
Since ovirt supports 250 hosts per dc, we can have a maximum of 
250*4KiB = ~ 1MiB.
Initialize 1MiB and allocate 128MiB to fit block storage extent size.

Bug-Url: https://bugzilla.redhat.com/2076200


roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
```diff
@@ -184,7 +184,7 @@
     args:
       warn: false
   - name: Initialize metadata volume
-    command: dd bs=1M count=1024 oflag=direct if=/dev/zero of="{{ he_metadata_disk_path }}"
+    command: fallocate -l 1G "{{ he_metadata_disk_path }}"
```
nirs, line 187: Is this a disk on a storage domain, so it can be file on NFS 3, NFS 4.2, Gluster
or a block device?

Here is fallocate support matrix:
- Glsuter: works
- NFS: works only with NFS 4.2. "fallocate --posix" works with any NFS version, but is extremely slow and problematic, vdsm never use it.
- Block:
  - "fallocatge --posix" does not work, but also does not fail (looks like a bug)
  - "fallocate --punch-hole" should work, but it can fail with EOPNOTSUPP
  - "fallocate --zero-range" should work, but it can fail with EOPNOTSUPP
  - "blkdiscard --zeroout" work with any block device (what vdsm uses)

Vdsm uses:
- File: /usr/libexec/vdsm/fallocate - helper using fallocate(2) or fall back to manual zeroing
- Block: "blkdiscard --zeroout --step 32m"

I think we need to understand first why do you need 1g. Most likely you need only few blocks and using dd is the right way to zero them.

arachmani: @tiraboschi do you have any idea why we initialize 1g?

roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
```diff
@@ -184,7 +184,7 @@
     args:
       warn: false
   - name: Initialize metadata volume
-    command: fallocate -l 1G "{{ he_metadata_disk_path }}"
+    command: dd bs=1M count=10 oflag=direct if=/dev/zero of="{{ he_metadata_disk_path }}"
```
nirs, line 187: Looks good - if we need only few MiB, there is no need to use fancy mechanism for zeroing, using dd is good.

roles/hosted_engine_setup/tasks/create_target_vm/01_create_target_hosted_engine_vm.yml
```diff
@@ -118,7 +118,7 @@
       - {
         name: 'he_metadata',
         description: 'Hosted-Engine metadata disk',
-        size: '1GiB',
+        size: '10MiB',
```
nirs, line 121: You will get 128 MiB disk on block storage because block storage use 128 MiB extent. So best use 128 MiB here.

nirs (commented):
The first commit is pointless since the second commit overwrite it.

I would close the bug as won't fix and use only the second commit.

sandrobonazzola (approved):
LGTM

nirs (approved):

roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
arachmani: Data is written at offset 4KiB*host_id, so we don't really need to initialize 1GiB, changed to 1MiB.

roles/hosted_engine_setup/tasks/create_target_vm/01_create_target_hosted_engine_vm.yml
arachmani: changed to 128MiB

PR #493, opened 2022-05-01
update iscsi parameters without overwriting the connection addresses

kobihk:
solve the issue as described in:
https://bugzilla.redhat.com/show_bug.cgi?id=2079896

### DIFF
--- a/roles/infra/roles/storages/tasks/main.yml
+++ b/roles/infra/roles/storages/tasks/main.yml
@@ -82,7 +82,7 @@
     auth: "{{ ovirt_auth }}"
     id: "{{ ansible_version.full is version('2.6.0', '>=') | ternary(item.1.id, item.1.id[0]) }}"
     storage: "{{ item.0.name }}"
-    address: "{{ storages[item.0.name][item.0.storage.type].address | default(omit) }}"
+    address: "{{ omit if item.0.storage.type == 'iscsi' or storages[item.0.name][item.0.storage.type].address | default(omit) }}"
     path: "{{ storages[item.0.name][item.0.storage.type].path | default(omit) }}"
     nfs_timeout: "{{ storages[item.0.name][item.0.storage.type].timeout | default(omit) }}"
     nfs_version: "{{ storages[item.0.name][item.0.storage.type].version | default(omit) }}"


arachmani:
Please add a changelog to make CI happy

roles/infra/roles/storages/tasks/main.yml
```diff
@@ -82,7 +82,7 @@
     auth: "{{ ovirt_auth }}"
     id: "{{ ansible_version.full is version('2.6.0', '>=') | ternary(item.1.id, item.1.id[0]) }}"
     storage: "{{ item.0.name }}"
-    address: "{{ storages[item.0.name][item.0.storage.type].address | default(omit) }}"
+    address: "{{ omit if item.0.storage.type == 'iscsi' or storages[item.0.name][item.0.storage.type].address | default(omit) }}"
```
michalskrivanek, line 85: why is this a correct fix?

ahadas: as I wrote elsewhere, I wouldn't go with this change - the other properties are also probably incorrect
we apply the same properties from `storages` to all connections which is wrong

kobihk: This is a patch to be able to keep running but as @ahadas said maybe we need to create a better solution.

IMHO when we create storage domains as we doing in lines:
https://github.com/oVirt/ovirt-ansible-collection/blob/master/roles/infra/roles/storages/tasks/main.yml#L6-L66
we should have the option to update the storage_connections in the same module "ovirt_storage_domain" 
so we will not have to run the "Update storage parameters" as we have it in:
https://github.com/oVirt/ovirt-ansible-collection/blob/master/roles/infra/roles/storages/tasks/main.yml#L80-L102
@mnecas wdyt?

mwperina: That seems to me like an RFE and not a regression

mwperina: We didn't do any relevant changes ovirt_storage_connection for at least 2 years

kobihk: @mwperina maybe it's more like a bug as you can see the full description in:
https://bugzilla.redhat.com/show_bug.cgi?id=2079896

ahadas: @mwperina you're right when looking at it from the ansible perspective but due to changes on the vdsm side, we may get to a worse situation then we would have in the past because of this issue

ahadas: @avihaie that's debatable - from QE perspective it's an awful regression because the tests fail
but from user's perspective it might be better to have the host set to non-operational and know about this issue (and ideally fix it)
however, in this context of ansible-collection, this is something we can fix in the next release

PR #497, opened 2022-05-10
Remove "Update storage parameters" play

roniezr:
This patch is a WA for the following issue:
https://bugzilla.redhat.com/show_bug.cgi?id=2079896

Jira ticket: RHV-45862


roniezr:
@mwperina FYI, WDYT?

nirs:
I guess that "WA" is "workaround". Lets keep the text free from
such shortcuts to make it more clear to everyone.

ahadas:
@roniezr thanks for posting that. that's what we had in mind but first we need to analyze from the storage point of view why it was added and what's the implication for dropping it

mnecas:
Resolved by https://github.com/oVirt/ovirt-ansible-collection/pull/500

PR #498, opened 2022-05-10
roles: hosted_engine_setup: Require 'detail: Up'

didib:
When checking, near the end of the deployment, if the engine is healthy,
do not only check that 'health' is 'good', but also that 'detail' is
'Up'.

We recently had a case in CI where due to some problem the engine took a
long time to start, timing out the monitoring in HA, causing it to stop
the VM, and 'detail' was 'Powering Down', but this went unnoticed,
causing things to fail much later. So, in such a case, fail early.

### DIFF
--- a/changelogs/fragments/498-require-detail-up.yml
+++ b/changelogs/fragments/498-require-detail-up.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+   - hosted_engine_setup - Require 'detail' to be 'Up' (https://github.com/oVirt/ovirt-ansible-collection/pull/498).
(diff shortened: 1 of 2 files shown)


changelogs/fragments/498-require-detail-up.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+   - hosted_engine_setup - Require 'detail: Up' (https://github.com/oVirt/ovirt-ansible-collection/pull/498).
```
arachmani, line 3: Looks like `:` is not allowed here

didib: Thanks :-(((

PR #500, opened 2022-05-23
role: infra: add warning for multiple storage connections

mnecas:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2079896

### DIFF
--- a/roles/infra/README.md
+++ b/roles/infra/README.md
@@ -197,6 +197,7 @@ More information about the parameters can be found in the [ovirt_host_network](h
 
 ### Storages
 To setup storages you can define dictionary variable called `storages`.
+This role supports only storages with one connection.
 The value of item in `storages` dictionary can contain following parameters (the key is always a name of the storage):
 
 | Name            | Default value  | Description                           |
--- a/roles/infra/roles/storages/tasks/main.yml
+++ b/roles/infra/roles/storages/tasks/main.yml
@@ -77,26 +77,31 @@
     - storages
     - storage_connections
 
+- name: Storage connection warning
+  debug:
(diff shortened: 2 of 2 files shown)


roles/infra/roles/storages/tasks/main.yml
```diff
@@ -77,26 +77,31 @@
     - storages
     - storage_connections
 
+- name: Storage connection warning
+  debug:
+    msg: "[WARNING]: The storage domain '{{ item.name }}' has multiple connections. This is not supported by the role."
```
michalskrivanek, line 82: also please say that it's skipped

mwperina (approved):
+1

ahadas:
> Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2079896

it doesn't fix it, maybe "tries to address" or "does the best we can for" :)

roles/infra/README.md
```diff
@@ -197,6 +197,7 @@ More information about the parameters can be found in the [ovirt_host_network](h
 
 ### Storages
 To setup storages you can define dictionary variable called `storages`.
+This role supports only storages with one connection.
```
kobihk, line 200: maybe you can add something like:
"In case of more than one connection, the storage connection update of this domain will be skipped."

kobihk (changes requested):

mnecas:
> > Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2079896
> 
> it doesn't fix it, maybe "tries to address" or "does the best we can for" :)

will update the comment :D

roles/infra/README.md
mnecas: +1

mnecas: +1

kobihk (approved):

PR #501, opened 2022-05-23
roles: hosted_engine_setup: fix archive ownership

michalskrivanek:
with umask 0777 (default for DISA STIG) the archive is not readable
later when we try to copy it as vdsm user.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2089332

### DIFF
--- a/changelogs/fragments/501-fix-archive-ownership.yml
+++ b/changelogs/fragments/501-fix-archive-ownership.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+     - hosted_engine_setup - fix archive ownership (https://github.com/oVirt/ovirt-ansible-collection/pull/501).
--- a/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
+++ b/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
@@ -133,6 +133,9 @@
     args:
       chdir: "{{ he_local_vm_dir }}"
       warn: false
+    become: true
+    become_user: vdsm
+    become_method: sudo
     changed_when: true
     tags: ['skip_ansible_lint']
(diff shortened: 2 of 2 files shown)


michalskrivanek:
verified with https://github.com/oVirt/ovirt-system-tests/pull/160

arachmani:
Please have a look at https://github.com/oVirt/ovirt-ansible-collection/pull/409
I started with changing the permissions and we decided to use `owner: vdsm` and `group: qemu` for the local VM dir, which worked fine on rhel-h with DISA STIG.
Following https://bugzilla.redhat.com/show_bug.cgi?id=2020620#c3, we need to support DISA STIG profile without setting umask 077 directly.

michalskrivanek:
> Please have a look at https://github.com/oVirt/ovirt-ansible-collection/pull/409
> 
> I started with changing the permissions and we decided to use `owner: vdsm` and `group: qemu` for the local VM dir, which worked fine on rhel-h with DISA STIG.

It couldn't have. The dir is fine, but the created file is not


> Following https://bugzilla.redhat.com/show_bug.cgi?id=2020620#c3, we need to support DISA STIG profile without setting umask 077 directly.

That's a misunderstanding. 077 is set by the profile. It needs to work with that umask. OST has been modified to do 022, that's why we didn't notice. I have no idea how QE passed though, either way now it doesn't:)

mwperina (approved):
+1

michalskrivanek:
or did you mean to create that tar as vdsm:qemu instead?

arachmani:
> or did you mean to create that tar as vdsm:qemu instead?

Yes, IMO it's better than setting 644 permissions

michalskrivanek:
verified in OST again

arachmani (approved):
LGTM

mnecas (approved):
LGTM

PR #503, opened 2022-05-26
Fix "'ansible' ModuleNotFoundError" in Disaster Recovery scripts

barpavel:
When executing any Disaster Recovery script, the following error is received:
```
[root@storage-ge-13 files]# ./ovirt-dr generate
  File "./ovirt-dr", line 15, in <module>
    import fail_back
  File "/usr/share/ansible/collections/ansible_collections/redhat/rhv/roles/disaster_recovery/files/fail_back.py", line 14, in <module>
    from ansible.module_utils.six.moves import input
ModuleNotFoundError: No module named 'ansible'
```

The reason for that is the following `import` statement: `from ansible.module_utils.six.moves import input`.

There is no need to use "`six`" library at all since Python2 is deprecated!
All "`import`"s of "`six`" module can be deleted now and the builtin Python3 "`input()`" function will be used instead.

Bug-Url: https://bugzilla.redhat.com/2090264

### DIFF
--- a/roles/disaster_recovery/files/ovirt-dr
+++ b/roles/disaster_recovery/files/ovirt-dr
@@ -10,7 +10,6 @@ import time
 import getopt
 
 from configparser import ConfigParser
-from six.moves import input
 
 import fail_back
 import fail_over
--- a/changelogs/fragments/503-fix-ansible-ModuleNotFoundError.yml
+++ b/changelogs/fragments/503-fix-ansible-ModuleNotFoundError.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - hosted_engine_setup - Fix "'ansible' ModuleNotFoundError" in Disaster Recovery scripts (https://github.com/oVirt/ovirt-ansible-collection/pull/503).
--- a/roles/disaster_recovery/files/fail_back.py
+++ b/roles/disaster_recovery/files/fail_back.py
@@ -11,7 +11,6 @@ import sys
 import time
 
 from configparser import ConfigParser
-from ansible.module_utils.six.moves import input
 
 from bcolors import bcolors
 
--- a/roles/disaster_recovery/files/fail_over.py
+++ b/roles/disaster_recovery/files/fail_over.py
@@ -11,7 +11,6 @@ import sys
 import time
 
 from configparser import ConfigParser
-from ansible.module_utils.six.moves import input
 
 from bcolors import bcolors
 
--- a/roles/disaster_recovery/files/generate_vars.py
+++ b/roles/disaster_recovery/files/generate_vars.py
@@ -9,7 +9,6 @@ import subprocess
 import sys
 
 from configparser import ConfigParser
-from ansible.module_utils.six.moves import input
 
 import ovirtsdk4 as sdk
 
--- a/roles/disaster_recovery/files/validator.py
+++ b/roles/disaster_recovery/files/validator.py
@@ -12,7 +12,6 @@ from ovirtsdk4 import types
 
 from bcolors import bcolors
 from configparser import ConfigParser
-from ansible.module_utils.six.moves import input
 
 
 INFO = bcolors.OKGREEN


michalskrivanek:
needs changelog entry, other than that lgtm

ahadas:
> AFAIU there is no need to use "`six`" library at all since Python2 is deprecated! All "`import`"s of "`six`" module can be deleted now and the builtin Python3 "`input`" function will be used instead.

@barpavel we may have 4.2 (or more probably 4.3) clusters that are based on EL7, where we used Python 2, wouldn't the DR be affected?

roles/disaster_recovery/files/ovirt-dr
```diff
@@ -123,7 +119,7 @@ def _get_log_conf(conf_file, log_file, log_level):
         log_level = settings.get(log_section, log_level_conf)
     else:
         log_level = "DEBUG"
-    return log_file, log_level
+    return conf_file, log_file, log_level
```
nirs, line 122: Why are you mixing this enhancement with a bug fix? It is best to keep it in a separate PR so we can merge the fix quickly.

The issue is not clear, maybe you need to show it in the commit message - how the interaction looks before and after the change.

barpavel: OK, I will separate it to 2 PRs after lunch, thanks.
The previous behavior is like that
```
[root@storage-ge-5 files]# ./ovirt-dr generate
Conf file 'dr.conf' does not exist. Please provide the configuration file location: dr1.conf           
Conf file 'dr.conf' does not exist. Please provide the configuration file location: dr1.conf
Log file: '/tmp/ovirt-dr-1653552400881.log'
[Generate Mapping File] Password is not initialized. Please provide the password for username admin@internal: 123456
...
```

barpavel:
> needs changelog entry, other than that lgtm

Will do.

nirs:
The real issue seems to be here:
https://github.com/oVirt/ovirt-ansible-collection/blob/master/roles/disaster_recovery/files/ovirt-dr

If this code should use ansible, it should use python3.8.

barpavel:
> from six.moves import input



> > AFAIU there is no need to use "`six`" library at all since Python2 is deprecated! All "`import`"s of "`six`" module can be deleted now and the builtin Python3 "`input`" function will be used instead.
> 
> @barpavel we may have 4.2 (or more probably 4.3) clusters that are based on EL7, where we used Python 2, wouldn't the DR be affected?

In that case (if we still use Python2) I can change the `import` from 
`from ansible.module_utils.six.moves import input`
to
`from six.moves import input`
as I described in the description.

roles/disaster_recovery/files/ovirt-dr
aesteve-rh: IMHO, I think it would be better to refactor the part that verifies that the config file exists and ask for the location if it doesn't out of `_get_log_conf`, which would always receive an existing file

barpavel:
> This part of the commit message is not helpful and best removed:
> 
> ```
> Executing:
>   pip3 install ansible
> solves the issue, but is not the right solution.
> 
> I compared the installed packages on the VM where the "pip3" command was executed
> (and by that fixed the issue) versus a VM where it was not executed (and the
> "ModuleNotFoundError" happens) and unfortunately failed to pinpoint what makes the difference.
> Also noticed that changing the import to:
>   from six.moves import input
> also solves the issue, thus "ansible.module_utils" prefix is the one that fails the flow.
> 
> I have no idea why the problem didn't happen in 4.4, but probably the investigation
> doesn't worth the effort: AFAIU there is no need to use "six" library at all since
> Python2 is deprecated!
> All "import"s of "six" module can be deleted now and the builtin Python3 "input"
> function will be used instead.
> ```
> 
> The issue seems to be using old python 2 compatibility code that is not needed in the last 2 years, and likely removed from ansible. Since we can use the builtin input module there is no need for the compatibility hack.
> 
> But what worries me is:
> 
> ```
>  ModuleNotFoundError: No module named 'ansible'
> ```
> 
> It can mean that you are using the wrong python version, which is a bigger problem.

Yes, strange, I see a lot of usages `from ansible.module_utils.XYZ` in `ovirt-ansible-collection`, so maybe indeed the issue is different and need to understand why QE env fails on that....

barpavel:
> The real issue seems to be here: https://github.com/oVirt/ovirt-ansible-collection/blob/master/roles/disaster_recovery/files/ovirt-dr
> 
> If this code should use ansible, it should use python3.8.

I see all DR files use:
`#!/usr/bin/python3`
While others (including those also having `from ansible.module_utils.XYZ`imports use:
`#!/usr/bin/python`.
Will check that soon (in 2-3 hours).

nirs:
The important question is if this script depends on ansible python package, or is is a standalone tool using other command line tools.

In the first case you want to use python3.8 since this is where ansible stuff is installed. In the second case using python3 is fine.

mwperina:
> The important question is if this script depends on ansible python package, or is is a standalone tool using other command line tools.
> 
> In the first case you want to use python3.8 since this is where ansible stuff is installed. In the second case using python3 is fine.

So it's not entirely true. AFAIR iif there is a python script executed from ansible playbook, then it's executed by the python installed on the target host and not the control host (the host where you execute the playbook from).

Of course you can force which python version you want to use using `command` directive and not relying on shebang, for example:

```
- name: Run python script
  ansible.builtin.command: /usr/bin/python38 /usr/share/mypackage/my_script.py
  register: mymotd
```

Just be aware that we are still supporting both EL7 and EL8 hosts (and hopefully soon also EL9 in oVirt), so you should choose only installed python version for each OS version (unless you want to install additional packages of course)

@mnecas Am I right?

barpavel:
I must admit that after all the comments from so many people - and I'm really thankful for your comments - I'm totally confused :(and have no clue what is the problem and what is the right solution and why it worked in 4.4 and stopped working.

@ahadas if we should support Python2, then probably need to keep on using `six` library for compatibility.

@nirs I see that on QE machines  `#!/usr/bin/python3` is replaced (probably during installation) with `/usr/libexec/platform-python` which is mapped to Python 3.6 and not 3.8: `/usr/libexec/platform-python -> ./platform-python3.6`. So it fails to find `ansible`, but after `pip3 install ansible` it does work with the same Python 3.6.8. How come?

@mnecas @mwperina maybe Martin Perina already provided the answer, but I didn't understand it, sorry then. I see there are lots of files having `from ansible.X.Y...` (for example in `ovirt-ansible-collection/plugins/modules`). Which makes me suspect that having `from ansible.module_utils` in the code is absolutely legal, so my solution to eliminate the redundant `six` library (or at least `ansible.module_utils` prefix) is not the right direction to fix the root cause of `ModuleNotFoundError: No module named 'ansible'` issue. Do you understand why it fails on QE env and why it started to happen now (worked in 4.4)?

And in regard to what @nirs wrote `If this code should use ansible, it should use python3.8`: is it possible to work in Python 3.6.8 (QE env) or Python2 (cluster versions 4.2. & 4.3 as @ahadas mentioned) while having `from ansible.X.Y...` in the code?

roles/disaster_recovery/files/ovirt-dr
barpavel: @aesteve-rh That's a good idea, will need to be a separate PR though :)
Also looks like `README.md` and `dr.conf` skeleton might also require a small update in regards to log filename and level for consistency between the 4 DR flows.

nirs:
I don't think we can support python 2 at this point. Users on RHEL 7 need to use the ovirt-dr script distributed for RHEL 7. Users on RHEL 8 can use only python 3.

The ovirt-dr script is not run by ansible - it runs ansible playbooks, so we cannot control
python version using ansible, it must be controlled by the #!/usr/bin/pythonXXX line.

Using the built in input function is the right way. Lets focus on this fix now and leave
the question of which python version to use for later.

barpavel:
> This part of the commit message is not helpful and best removed

Removed the redundant description from both the commit and PR messages.

barpavel:
> needs changelog entry, other than that lgtm

Done. Added the changelog entry according to examples from other PRs.

nirs (approved):

ahadas:
@barpavel ok, great, thanks for checking

ahadas (approved):

michalskrivanek (approved):

mnecas (approved):

PR #504, opened 2022-05-26
roles: hosted_engine_setup: final: Collect engine logs

didib:
Collect logs also on failures in 03_hosted_engine_final_tasks.yml.

The logs are still only from the local VM's disk, not the target.

Pushing this patch because we had cases where 'Check OVF_STORE volume
status' timed out.


arachmani (commented):
LGTM, changelog is missing.

arachmani (approved):

PR #506, opened 2022-05-26
roles: hosted_engine_setup: Force updating the OVF store

didib:
Directly ask the engine to update the OVF store instead of just hoping
that it eventually does.

If this patch works, we might remove other code, such as
setting/restoring OvfUpdateIntervalInMinutes and related.

Related-To: https://bugzilla.redhat.com/2090638

### DIFF
--- a/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
+++ b/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
@@ -9,6 +9,13 @@
       description: "Hosted engine VM"
       serial_console: true
       auth: "{{ ovirt_auth }}"
+  - name: Force updating the OVF store
+    ovirt_storage_domain:
+      auth: "{{ ovirt_auth }}"
+      name: "{{ he_storage_domain_name }}"
+      state: update_ovf_store
+    retries: 12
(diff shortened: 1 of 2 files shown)


roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
```diff
@@ -9,6 +9,12 @@
       description: "Hosted engine VM"
       serial_console: true
       auth: "{{ ovirt_auth }}"
+  - name: Force updating the OVF store
+    ovirt_storage_domain:
+      name: {{ he_storage_domain_name }}
+      state: update_ovf_store
+    retries: 12
+    delay: 10
```
nirs, line 17: Looks promising.

arachmani:
@didib, should it fix https://bugzilla.redhat.com/show_bug.cgi?id=2090638?

didib:
> @didib, should it fix https://bugzilla.redhat.com/show_bug.cgi?id=2090638?

It might help there, yes, but still not sure about the root cause there - if it's a storage issue, it won't fix it.

didib:
I don't mind adding a Bug-Url if that's the question.

didib:
Updated the commit message, nothing else.

PR #508, opened 2022-06-01
roles: hosted_engine_setup: Use command instead of firewalld module

arachmani:
Bug-Url: https://bugzilla.redhat.com/2092229

### DIFF
--- a/changelogs/fragments/508-ovirt_hosted_engine_setup-use-firewalld-command.yml
+++ b/changelogs/fragments/508-ovirt_hosted_engine_setup-use-firewalld-command.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - hosted_engine_setup - Use command instead of firewalld module (https://github.com/oVirt/ovirt-ansible-collection/pull/508).
--- a/roles/hosted_engine_setup/tasks/bootstrap_local_vm/05_add_host.yml
+++ b/roles/hosted_engine_setup/tasks/bootstrap_local_vm/05_add_host.yml
@@ -10,11 +10,7 @@
     retries: 30
     delay: 20
   - name: Open a port on firewalld
-    firewalld:
-      port: "{{ he_webui_forward_port }}/tcp"
-      permanent: false
-      immediate: true
-      state: enabled
+    command: firewall-cmd --zone=public --add-port {{ he_webui_forward_port }}/tcp
   - name: Expose engine VM webui over a local port via ssh port forwarding
     command: >-
       sshpass -e ssh -tt -o ServerAliveInterval=5 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -g -L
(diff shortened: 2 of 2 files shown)


michalskrivanek:
OST failed with
11:11:55 E [ INFO ] TASK [ovirt.ovirt.hosted_engine_setup : Force updating the OVF store]

11:11:55 E [ ERROR ] ovirtsdk4.Error: Fault reason is "Operation Failed". Fault detail is "[Cannot ${action} Storage. Storage Domain OVF data is currently being updated.]". HTTP response code is 409.

11:11:55 E [ ERROR ] fatal: [localhost]: FAILED! =&gt; {"changed": false, "msg": "Fault reason is \"Operation Failed\". Fault detail is \"[Cannot ${action} Storage. Storage Domain OVF data is currently being updated.]\". HTTP response code is 409."}



due to recent change that was imho unneeded..."If it ain't broke, don't fix it" ;)

michalskrivanek:
..so now you have to fix it. @arachmani @didib I would perhaps suggest to revert and open similar PR again and wait for someone to implement it in a more robust way with the desired  cleanup. wdyt?

michalskrivanek (approved):

arachmani:
> -1 Please describe the gluster issue before moving from firewalld module to command, maybe just gluster invocation needs to be changed ...

AFACT we have encountered this because there is no firewall module for python38 while ansible-controller uses python38.
I tried to figure out how come it works for RHV QE and not for RHHI-V QE but unfortunately didn't get to a conclusion.
RHHI-V QE fails to install HE role on both remote and local host while RHV QE is able to install on remote though doesn't run the role locally.
I tried executing the HE role directly on a local host and failed with the same error, note that `hosted-engine --deploy` works fine on a local host.
The solution for that is to use the `command` module or add `-e ansible_python_interpreter=/usr/bin/python3` when executing the role directly.

arachmani:
> So one more question: I've just noticed https://bugzilla.redhat.com/show_bug.cgi?id=2041997 What optoin are we using for HE installation on locahost? Are we using the new `ansible_connection=local` or `ansible-playbook -c local` or are we using the old way from ansible <= 2.9

We use this yml file as an example when running the role directly:
https://github.com/oVirt/ovirt-ansible-collection/blob/bbee76a9a0ca313f461b58d4f0785f99c3e45254/roles/hosted_engine_setup/examples/hosted_engine_deploy_localhost.yml#L4

And a similar yml file when using `hosted-engine --deploy`:
https://github.com/oVirt/ovirt-hosted-engine-setup/blob/master/src/he_ansible/trigger_role.yml

I tried to use `firewalld` module on localhost along with `ansible-playbook -c local`, but got the same error.

mwperina (approved):
OK, so LGTM

mnecas (approved):
LGTM

PR #509, opened 2022-06-01
roles: hosted_engine_setup: Fix keycloak activation/checking

didib:
keycloak integration had two env keys for this, and recently unified
them. We used the one that was removed. So fix, to use the current one.


arachmani (approved):
LGTM

PR #510, opened 2022-06-01
Cluster upgrade: fix wait_condition

mnecas:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2036613

### DIFF
--- a/plugins/modules/ovirt_host.py
+++ b/plugins/modules/ovirt_host.py
@@ -613,7 +613,16 @@ def main():
             ret = hosts_module.action(
                 action='upgrade',
                 action_condition=lambda h: h.update_available,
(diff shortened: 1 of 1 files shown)


dupondje:
It seems like the event 840 is emitted before the host even goes into maintenance mode.
So this does not seem to fix the issue :(

Guess we will have to wait on one of the following events:
    HOST_UPGRADE_FAILED(841, AuditLogSeverity.ERROR),
    HOST_UPGRADE_FINISHED(842),
    HOST_UPGRADE_FINISHED_AND_WILL_BE_REBOOTED(888),
    HOST_UPGRADE_FINISHED_MANUAL_HA(890, AuditLogSeverity.WARNING),

PR #511, opened 2022-06-02
vars files added for ovirt 4.5

kobihk:


mnecas (changes requested):
These vars should be only for downstream so no 4.5.
And there is no difference between 4.4 and 4.5 repos.
Don't see the reason for this patch/.
-1

kobihk:
@mnecas we use this role also with downstream but np I"ll fix it on our code
Thanks!!

PR #512, opened 2022-06-03
Revert "roles: hosted_engine_setup: Force updating the OVF store"

michalskrivanek:
Reverts oVirt/ovirt-ansible-collection#506

it seems to be flaky, unfortunately. OST sometimes fails with
11:11:55 E [ INFO ] TASK [ovirt.ovirt.hosted_engine_setup : Force updating the OVF store]

11:11:55 E [ ERROR ] ovirtsdk4.Error: Fault reason is "Operation Failed". Fault detail is "[Cannot ${action} Storage. Storage Domain OVF data is currently being updated.]". HTTP response code is 409.

11:11:55 E [ ERROR ] fatal: [localhost]: FAILED! => {"changed": false, "msg": "Fault reason is "Operation Failed". Fault detail is "[Cannot ${action} Storage. Storage Domain OVF data is currently being updated.]". HTTP response code is 409."}

### DIFF
--- a/changelogs/fragments/506-force-updating-ovf-store.yml
+++ b/changelogs/fragments/506-force-updating-ovf-store.yml
@@ -1,3 +1,0 @@
----
-bugfixes:
-  - hosted_engine_setup - Force updating the ovf store (https://github.com/oVirt/ovirt-ansible-collection/pull/506).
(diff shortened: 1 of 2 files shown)


michalskrivanek:
hm, it seems to fail fairly often actually, and it was hidden due to keycloak issues for a while. one more try on master but if it fails 3/3 i'll revert right away....

Issue #491, opened 2022-04-26
Use python 3.8 in automation


didib:
Based on the errors I got, for which I pushed #490, it seems like we use python 3.6 in automation, despite our code being intended to be run on python 3.8 (with ansible-core 2.12).

We should run at least all sanity tests in python 3.8.

Part of this should probably be to revert #490.

I am creating this issue after talking with @mwperina and he mentioned a few other related things, including use of ansible-lint (of correct version(s)), need to test also on 3.9 for compatibility of future ansible-core, etc. - not adding more details because I do not have the full picture.

PR #518, opened 2022-06-06
CI: Use python38 on el8 with ansible-core 2.12 and python39 on el9 with ansible-core 2.13 

mnecas:
Fixes: https://github.com/oVirt/ovirt-ansible-collection/issues/491

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -27,18 +27,15 @@ jobs:
       - name: Update pip
         run: pip3 install pip --upgrade
 
-      - name: Install rpm ansible-core
-        run: yum install ansible-core -y
-
-      - name: Install pip modules
-        if: ${{ matrix.shortcut == 'cs9' }}
-        run: >
-          pip-3 install pycodestyle pylint==2.4.4 voluptuous yamllint "rstcheck<3.5.0" antsibull-changelog "rich<11.0.0" "ansible-lint<5.0.0" 'cryptography<37.0.0'
-
-      - name: Install pip modules
+      - name: Install python3.8 modules
         if: ${{ matrix.shortcut == 'cs8' }}
         run: >
           pip-3.8 install pycodestyle pylint==2.4.4 voluptuous yamllint "rstcheck<3.5.0" antsibull-changelog "rich<11.0.0" "ansible-lint<5.0.0" 'cryptography<37.0.0'
+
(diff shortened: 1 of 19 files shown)


.github/workflows/build.yaml
```diff
@@ -27,9 +27,20 @@ jobs:
       - name: Update pip
         run: pip3 install pip --upgrade
```
mwperina, line 28: So this is relevant ony for CS9, right?

.github/workflows/build.yaml
```diff
@@ -27,9 +27,20 @@ jobs:
...
 
-      - name: Install pip modules
+      - name: Install python3.8 modules
+        if: ${{ matrix.shortcut == 'cs8' }}
         run: >
-          pip3 install pycodestyle pylint==2.4.4 voluptuous yamllint "rstcheck<3.5.0" antsibull-changelog "rich<11.0.0" "ansible-lint<5.0.0" 'cryptography<37.0.0'
+          pip-3.8 install pycodestyle pylint==2.4.4 voluptuous yamllint "rstcheck<3.5.0" antsibull-changelog "rich<11.0.0" "ansible-lint<5.0.0" 'cryptography<37.0.0'
```
mwperina, line 33: Also once we moved to Python 3.8, do we really need those version limitation around cryptography, rstcheck, ...?

mwperina (changes requested):

.github/workflows/build.yaml
mnecas: good point, will remove it

.github/workflows/build.yaml
mnecas: I still think so, I know there were some pip installation problems without it.
I'll create one testing commit without CS9

mnecas: Looks like it was related only to python36, both passed the installation even without the upgrade.
Removing

nirs:
So now we need to build ovirt-imageio for python 3.9 in the centos 9 build?

mnecas:
> So now we need to build ovirt-imageio for python 3.9 in the centos 9 build?

the default python on centos 9 stream is python 3.9, so if you are building the ovirt-imageio for the el9 should be fine.
You don't need to do that python versioning magic as on centos 8 stream.

mwperina (approved):
+1

PR #516, opened 2022-06-06
Remove deprecated python package distutils

mnecas:
Fixes part of https://github.com/oVirt/ovirt-ansible-collection/issues/495

PR #522, opened 2022-06-06
Revert "Remove deprecated python package distutils"

mnecas:
Reverts oVirt/ovirt-ansible-collection#516


mnecas:
With 516 we would need to add a new dependency a rebuild the package for python38.

PR #534, opened 2022-06-13
ovirt_storage_domain: Fix error for inaccessible storage domain

mnecas:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2084235

Now when running playbook like:
```yml
# Add data NFS storage domain
    - ovirt.ovirt.ovirt_storage_domain:
        name: Ble        #This name is used in the different data center
        host: host_mixed_2
        data_center: golden_env_mixed
        nfs:
          address: "{{nfs_server}}"
          path: "{{nfs_path}}"
        auth: "{{ovirt_auth}}"
```

It will throw exception from engine:
```
    "msg": "Fault reason is \"Operation Failed\". Fault detail is \"[Cannot attach Storage. The relevant Storage Domain is inaccessible.\n-Please handle Storage Domain issues and retry the operation.]\". HTTP response code is 409."
```


mwperina (approved):
+1

PR #280, opened 2021-05-28
fix ansible-lint for basic roles

mnecas:
Fixes: https://github.com/oVirt/ovirt-ansible-collection/issues/265

PR #536, opened 2022-06-13
CI: Run ansible-lint 4

mnecas:
The https://github.com/oVirt/ovirt-ansible-collection/pull/280 bumped ansible-lint to 6.0.0 but with that, we needed to exclude not fixed roles, so temporarily I'm adding the ansible-lint 4 for all the roles.


mwperina (approved):
+1

PR #540, opened 2022-06-15
Make hosted engine VM headless

ahadas:


ahadas:
@michalskrivanek wouldn't it be as simple as that?

arachmani:
In that case, you also want to remove this task: https://github.com/oVirt/ovirt-ansible-collection/blob/0796b09a3fe36c83f2f0e3a862b2ef7a0aaab5b5/roles/hosted_engine_setup/tasks/create_target_vm/01_create_target_hosted_engine_vm.yml#L133:

didib:
What's the status of this? Can we have some background? Bug-Url?

This was recently discussed somewhere, and IIRC @michalskrivanek said some people still do want a graphical console, for various reasons.

Personally I do not have a strong opinion either way. I only use the serial console, if available, but understand if people want the graphical one.

~ 12 or so years ago, on then-current hardware/software, I compared the cpu load of a qemu process on the host. With a graphical console (showing just a text 'login:'), it was ~ 6% of one cpu core, for an idle VM. With a serial-only one, it was less than 1%.

ahadas:
@sandrobonazzola @didib it was posted as an alternative for changing the video device of the hosted-engine from qxl to vga, so it could run on el9. but the latter was already implemented so we no longer need this change for this reason and it's unlikely that I'll be able to follow up on this, therefore closing

PR #543, opened 2022-06-16
Fix 'ansible-lint' warnings for 'disaster_recovery' and 'remove_stale_lun' roles.

barpavel:
Fix `ansible-lint 6.0.0` for `disaster_recovery` and `remove_stale_lun` roles.

**WIP: many more commits on the way...**

Bug-Url: https://bugzilla.redhat.com/2097332

PR #545, opened 2022-06-22
Fix lint errors he setup

didib:


automation/build.sh
```diff
@@ -68,7 +68,7 @@ cd "$COLLECTION_DIR"
 
 ansible-test sanity
 antsibull-changelog lint -v
-ansible-lint roles/* --exclude roles/hosted_engine_setup --exclude roles/disaster_recovery --exclude roles/remove_stale_lun -x experimental
+ansible-lint roles/* --exclude --exclude roles/disaster_recovery --exclude roles/remove_stale_lun -x experimental
```
mwperina, line 71: Shouldn't also the empty `--exclude` option be removed?

plugins/modules/ovirt_event.py
```diff
@@ -76,7 +76,7 @@
         description:
             - "The id of the template associated with this event."
         type: str
-    user:
+    ansible.builtin.user:
```
mwperina, line 79: How is this related to ansible? Isn't `user` just an id of a user associated within this event and as such it has no relation to https://docs.ansible.com/ansible/latest/collections/ansible/builtin/user_module.html ,right?

roles/disaster_recovery/files/generate_vars.py
```diff
@@ -146,7 +146,7 @@ def _validate_connection(self,
             msg = "Connection to setup has failed. " \
                   "Please check your credentials: " \
                   "\n URL: " + url + \
-                  "\n user: " + username + \
+                  "\n ansible.builtin.user: " + username + \
```
mwperina, line 149: Why to replace that in a message string?

roles/disaster_recovery/files/validator.py
```diff
@@ -117,10 +117,10 @@ def _set_dr_conf_variables(self, conf_file):
         self.var_file = var_file
 
         self.primary_pwd = input(
-            "%s%sPlease provide password for the primary setup: %s"
+            "%s%sPlease provide password for the primary ansible.builtin.setup: %s"
```
mwperina, line 120: Why to replace that in a message string?

roles/disaster_recovery/files/validator.py
```diff
@@ -117,10 +117,10 @@ def _set_dr_conf_variables(self, conf_file):
...
             % (INPUT, PREFIX, END))
         self.second_pwd = input(
-            "%s%sPlease provide password for the secondary setup: %s"
+            "%s%sPlease provide password for the secondary ansible.builtin.setup: %s"
```
mwperina, line 123: Why to replace that in a message string?

roles/disaster_recovery/files/validator.py
```diff
@@ -689,7 +689,7 @@ def _validate_connection(self, url, username, password, ca):
                 "%s%sConnection to setup has failed."
                 " Please check your credentials: "
                 "\n%s URL: %s"
-                "\n%s user: %s"
+                "\n%s ansible.builtin.user: %s"
```
mwperina, line 692: Why to replace that in a message string?

roles/hosted_engine_setup/hooks/after_setup/add_host_storage_domain.yml
```diff
@@ -1,15 +1,15 @@
...
   loop_control:
     loop_var: file_item
-- debug: var=gluster
+- ansible.builtin.debug: var=gluster
 
 - name: Set Engine public key as authorized key without validating the TLS/SSL certificates
   connection: ssh
   authorized_key:
-    user: root
+    ansible.builtin.user: root
```
mwperina, line 12: It's a `user` parameter and not a call to `ansible.builtin.user` module, right?

mwperina (changes requested):

plugins/modules/ovirt_event.py
mnecas: removing this resolves https://github.com/oVirt/ovirt-ansible-collection/issues/546

automation/build.sh
didib: Yes...

plugins/modules/ovirt_event.py
didib: Sorry, too broad automatic change. Same for others below.

roles/hosted_engine_setup/hooks/after_setup/add_host_storage_domain.yml
didib: Yes. I wonder if there is a simple way to find out only user module calls. Will check/fix manually for now.

didib:
> If you run an automatic script for all roles please sync with @barpavel because you are also changing the disaster_recovery role or remove the patch.

I'll remove, for now.

didib:
In the two commits where I did 'find . ... | xargs sed' I changed them to 'find roles/hosted_engine_setup'. We can obviously use these `find` commands with other sub-paths if/when needed later.

plugins/modules/ovirt_event.py
didib: > removing this resolves #546

Ok, I'll try to rebase on a revert for it.

roles/hosted_engine_setup/templates/user-data.j2
```diff
@@ -14,7 +14,7 @@ chpasswd:
 timezone: {{ he_time_zone }}
 {% endif %}
 bootcmd:
-  - if grep -Gq "^\s*PermitRootLogin" /etc/ssh/sshd_config; then sed -re "s/^\s*(PermitRootLogin)\s+(yes|no|without-password)/\1 yes/" -i.$(date -u +%Y%m%d%H%M%S) /etc/ssh/sshd_config; else echo "PermitRootLogin yes" >> /etc/ssh/sshd_config; fi
-  - if grep -Gq "^\s*UseDNS" /etc/ssh/sshd_config; then sed -re "s/^\s*(UseDNS)\s+(yes|no)/\1 no/" -i.$(date -u +%Y%m%d%H%M%S) /etc/ssh/sshd_config; else echo "UseDNS no" >> /etc/ssh/sshd_config; fi
+  - if grep -Gq "^\s*PermitRootLogin" /etc/ssh/sshd_config; then sed -re "s/^\s*(PermitRootLogin)\s+(yes | no | without-password)/\1 yes/" -i.$(date -u +%Y%m%d%H%M%S) /etc/ssh/sshd_config; else echo "PermitRootLogin yes" >> /etc/ssh/sshd_config; fi
+  - if grep -Gq "^\s*UseDNS" /etc/ssh/sshd_config; then sed -re "s/^\s*(UseDNS)\s+(yes | no)/\1 no/" -i.$(date -u +%Y%m%d%H%M%S) /etc/ssh/sshd_config; else echo "UseDNS no" >> /etc/ssh/sshd_config; fi
```
mnecas, line 18: watch out here, not sure if it will affect the code but want to just point it out

roles/hosted_engine_setup/templates/broker.conf.j2
```diff
@@ -5,4 +5,4 @@ source-email = {{ he_source_email }}
 destination-emails = {{ he_dest_email }}
 
 [notify]
-state_transition = maintenance|start|stop|migrate|up|down
+state_transition = maintenance | start | stop | migrate | up | down
```
mnecas, line 8: same here not sure if it will affect the code

roles/hosted_engine_setup/tasks/initial_clean.yml
```diff
@@ -3,15 +3,15 @@
...
       name: libvirtd
       state: stopped
       enabled: true
   - name: Drop vdsm config statements
-    shell: >-
+    ansible.builtin.shell: >-
       [ -r {{ item }} ] && sed -i
       '/## beginning of configuration section by
-      vdsm-4.[0-9]\+.[0-9]\+/,/## end of configuration section by vdsm-4.[0-9]\+.[0-9]\+/d' {{ item }} || :
+      vdsm-4.[0-9]\+.[0-9]\+/,/## end of configuration section by vdsm-4.[0-9]\+.[0-9]\+/d' {{ item }} | | :
```
mnecas, line 14: same here

mnecas (commented):
noticed some changes outside yml file

arachmani (commented):
Please add bug-url

roles/hosted_engine_setup/examples/required_networks_fix.yml
```diff
@@ -38,19 +38,19 @@
...
+        auth: "{{ ovirt_auth }}"
+        name: myhost
+        interface: eth0
+        networks:
+          - name: net1
+    - name: Activate host myhost
+      ovirt.ovirt.ovirt_host:
+        auth: "{{ ovirt_auth }}"
+        name: myhost
+        state: present
```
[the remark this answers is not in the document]
didib: > Not sure if the spaces here are required,

They are. I don't mind adding the test that failed without this (don't remember anymore, can check).

> if so, shouldn't use double spaces?

AFAIU it wasn't due to too-short indents, but simply incorrect. Not sure how it worked before the patch. Please note that this is an example, not used by the code. I did test it when adding it, quite some time ago.

didib:
> Please add bug-url

Done

automation/build.sh
```diff
@@ -69,6 +69,7 @@ cd "$COLLECTION_DIR"
 ansible-test sanity
 antsibull-changelog lint -v
 ansible-lint roles/* --exclude roles/hosted_engine_setup --exclude roles/disaster_recovery --exclude roles/remove_stale_lun -x experimental
+ansible-lint -x var-spacing -x ignore-errors -x var-naming roles/hosted_engine_setup
```
mwperina, line 72: Shouldn't `--exclude roles/disaster_recovery --exclude roles/remove_stale_lun` stay there as they should be fixed in https://github.com/oVirt/ovirt-ansible-collection/pull/543 ?

mwperina (changes requested):

automation/build.sh
didib: I didn't change the line (71), only added a new one. I want specific exclusions for HE (as long as we need them). The new line only runs on HE, not on roles/* .

mnecas: I'm okay with skipping `var-naming` these changes could break backward compatibility.
`ignore-errors` please check out the article about `ignore_error` vs `changed_when` [1] but I'm okay with skipping this one if you have a reason that you want red ignored errors.
and `var-spacing` was already discussed 
[1] https://medium.com/@sbarnea/why-ansible-ignore-errors-is-evil-500fb6e81229

didib: I now read it, thanks for the pointer.

1. I do not completely disagree, just claim that going through each and every use of ignore_errors and spending the time investigating the exact reasoning, deciding on the best approach, etc., takes a lot of time, and I do not think it's mandatory to do for upgrading lint.

2. Quickly replacing every use with failed_when, IMO, is useless. At most, it makes the output slightly "nicer".

3. It's also not working, see above. Didn't investigate carefully. It seems like 'failed_when' can't be done for a block, where 'ignore_errors' apparently can be.

4. Most importantly: The reasoning of [1] applied also when we _added_ these ignore_errors. We should have been convinced at the time that they are ok. I now looked again at a few of them, and still think they make sense. Take e.g. "Check if keycloak is configured". The result is registered in keycloak_configured_out, which is used immediately later. So if the command failed in a way that will make the next task fail too, we'd notice, without further confusion, despite the use of ignore_errors. Similarly for "Wait for the host to be up". Some we really simply do not care about, e.g. "Generate the error message from the engine events" - this one is ran only if we already failed, and is only for outputting more information to the user. If this fails, ok - the user will have to spend some more time checking logs. No further harm here.

didib:
Last patch I added, "he: Add loop variables to include_tasks uses", is not really related to current PR, as lint does not warn about it. But:
1. I originally thought it is, and also commented about this in the bug
2. I think it should. This definitely sounds to me like a more important issue than some of the things it does check.

didib:
Rebased

mnecas (approved):
LGTM

didib:
Documenting here a long discussion with @mnecas. With the current version of this PR, ansible-lint outputs (to stderr):
```
[DEPRECATION WARNING]: Use 'ansible.utils.ipaddr' module instead. This feature will be removed from ansible.netcommon in a 
release after 2024-01-01. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: Use 'ansible.utils.ipv4' module instead. This feature will be removed from ansible.netcommon in a 
release after 2024-01-01. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: Use 'ansible.utils.ipv6' module instead. This feature will be removed from ansible.netcommon in a 
release after 2024-01-01. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```
This is a result of using filters ipaddr, ipv4 and ipv6, without a namespace. Right now, they are used from ansible.netcommon, which outputs these warnings. I tried fixing this (with [1], and then [2]), but this does not work, because right now, the version of ansible-collection-ansible-utils that we use (which we get from CentOS Stream 8), does not include these filters.

So for now, just giving up on this. Later, once they are included in ansible-utils, we can simply remove 'Requires: ansible-collection-ansible-utils', and will automatically (hopefully) get the version from utils and get rid of the warning.

[1] https://github.com/oVirt/ovirt-ansible-collection/commit/bece1c3b1f347b576367d1298cf03287207afa0a lint: he: Fix deprecation warnings from ansible.netcommon

[2] https://github.com/oVirt/ovirt-ansible-collection/commit/579422d6d18e591829f883c94b1ed36a378fbd95 lint: he: Use ansible.utils collection

didib:
Failed. I'll fix next week

automation/build.sh
barpavel: This discussion is relevant for me as well :) in my [#543](https://github.com/oVirt/ovirt-ansible-collection/pull/543).
So is there a bottom line decision?
Should I skip `ignore_errors` or replace them with `failed_when`?

mnecas: yeah, I'll add `.ansible-lint` file for default ignores

didib:
(Just rebased, still didn't fix)

arachmani (approved):
LGTM

mwperina:
Unfortunately there is still some sanity error on EL9:
```
ERROR: Command "ansible-doc -t callback ovirt.ovirt.stdout" returned exit status 1.
>>> Standard Error
  File "/usr/bin/ansible-doc", line 33, in <module>
    sys.exit(load_entry_point('ansible-core==2.13.1', 'console_scripts', 'ansible-doc')())
  File "/usr/bin/ansible-doc", line 25, in importlib_load_entry_point
    return next(matches).load()
```

didib:
I think it already passed previously, so it's a new issue. Is anyone looking at it?

roles/hosted_engine_setup/tasks/add_engine_as_ansible_host.yml
```diff
@@ -7,6 +7,7 @@
     become: false
     ansible.builtin.command: whoami
     register: username_on_host
+    changed_when: true
```
barpavel, line 10: Shouldn't it be `false`?
From some reading I did about `changed_when`, all the read-only `shell` & `command` tasks should declare `changed_when: false` and the other might even need to check the return status to be sure whether the task eventually modified something or not.
And `whoami` looks like very much read-only.
Am I right? Since I'm also fixing same kind of failures.
Same comment might be relevant for other places as well.

roles/hosted_engine_setup/tasks/search_available_network_subnet.yaml
```diff
@@ -29,24 +30,26 @@
...
     - name: Define 3rd chunk
-      set_fact:
+      ansible.builtin.set_fact:
         chunk: 1000
       when: chunk is not defined
     - name: Set 3rd chunk
-      set_fact:
+      ansible.builtin.set_fact:
         chunk: "{{ chunk|int + 45 }}"  # 200 tries
+      changed_when: true
```
barpavel, line 39: Needed here (and in some more `set_fact`)?
It's not `shell` or `command`...

roles/hosted_engine_setup/tasks/add_engine_as_ansible_host.yml
arachmani: AFAICT, it doesn't really matter, but yes, `changed_when: false` make more sense.

roles/hosted_engine_setup/tasks/search_available_network_subnet.yaml
didib: Thanks. Compared with current code, the next push:
- Removes all 'changed_when:' lines I added that were not reported by ansible-lint. Not sure why I added them originally, will ignore for now.
- Changes a few to 'false', when it makes sense.

sandrobonazzola (approved):

mnecas (approved):

Issue #546, opened 2022-06-22
CI fails validate-modules in plugins/modules/ovirt_event.py


didib:
2022-06-22T06:04:41.0629940Z ERROR: Found 3 validate-modules issue(s) which need to be resolved:
2022-06-22T06:04:41.0630645Z ERROR: plugins/modules/ovirt_event.py:0:0: doc-missing-type: Argument 'user' in argument_spec uses default type ('str') but documentation doesn't define type
2022-06-22T06:04:41.0631442Z ERROR: plugins/modules/ovirt_event.py:0:0: nonexistent-parameter-documented: Argument 'ansible.builtin.user' is listed in DOCUMENTATION.options, but not accepted by the module argument_spec
2022-06-22T06:04:41.0632192Z ERROR: plugins/modules/ovirt_event.py:0:0: undocumented-parameter: Argument 'user' is listed in the argument_spec, but not documented in the module documentation

didib:
This failed #545, I'll try to fix

mnecas:
This issue is caused by changing the `user` in docs to `ansible.builtin.user` in your patch.

PR #547, opened 2022-06-22
Skip validate-module for plugins/modules/ovirt_event.py

didib:
Fails with:

https://github.com/oVirt/ovirt-ansible-collection/issues/546


arachmani (approved):
LGTM

PR #550, opened 2022-06-22
ovirt_snapshot: Add vm_id

mnecas:
Issue: When we have multiple VMs with the same name we were not able to specify which to select.


snecklifter (approved):
LGTM

PR #557, opened 2022-06-29
he: Handle migration to hosts that use systemd-coredump

tinez:
This PR adds proper handling of a situation where the HE is migrated
from a host that was using abrt to a host that's using systemd-coredump.

Bug-Url: https://bugzilla.redhat.com/2101481
Signed-off-by: Marcin Sobczyk <msobczyk@redhat.com>

### DIFF
--- a/roles/hosted_engine_setup/tasks/initial_clean.yml
+++ b/roles/hosted_engine_setup/tasks/initial_clean.yml
@@ -29,10 +29,14 @@
     args:
       warn: false
     environment: "{{ he_cmd_lang }}"
-  - name: Verify abrt-action-save-package-data file
+  - name: Check if vdsm's abrt-action-save-package-data config exists
     stat:
       path: /etc/abrt/abrt-action-save-package-data.conf
-    register: abrt_exists
+    register: abrt_vdsm_config
+  - name: Check if abrt is installed
+    stat:
+      path: /usr/share/abrt/conf.d/abrt-action-save-package-data.conf
+    register: abrt_installed_config
   - name: Restore initial abrt config files
     copy:
       remote_src: true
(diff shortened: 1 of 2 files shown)


arachmani (approved):
LGTM

roles/hosted_engine_setup/tasks/initial_clean.yml
```diff
@@ -56,12 +60,22 @@
         src: /usr/share/abrt/conf.d/plugins/vmcore.conf,
         dest: /etc/abrt/plugins/vmcore.conf
       }
-    when: abrt_exists.stat.exists
+    when: abrt_vdsm_config.stat.exists and abrt_installed_config.stat.exists
```
mwperina, line 63: Please use "more readable" :-) syntax:
```
    when:
      - abrt_vdsm_config.stat.exists
      - abrt_installed_config.stat.exists
```

roles/hosted_engine_setup/tasks/initial_clean.yml
```diff
@@ -56,12 +60,22 @@
...
+  - name: Remove vdsm's abrt config files
+    file:
+      state: absent
+      path: "{{ item }}"
+    with_items:
+      - /etc/abrt/abrt-action-save-package-data.conf
+      - /etc/abrt/abrt.conf
+      - /etc/abrt/plugins/CCpp.conf
+      - /etc/abrt/plugins/vmcore.conf
+    when: abrt_vdsm_config.stat.exists and not abrt_installed_config.stat.exists
```
mwperina, line 78: Wouldn't it be better to encapsulate `Restore initial abrt config files`, `Restart abrtd service`  and `Remove vdsm's abrt config files` into a `block` and use the condition for the whole block?

mwperina (changes requested):

roles/hosted_engine_setup/tasks/initial_clean.yml
tinez: `Remove vdsm's...` uses a different condition - notice the `not`, so the block would include only the first two. Do you think it's still worth it?

mwperina: Ahh sorry, missed that. So the I'd say stay as is, just please reformat the conditions

roles/hosted_engine_setup/tasks/initial_clean.yml
tinez: Done, hope the one with `not` will work too

mwperina (approved):
+1

PR #559, opened 2022-07-05
Improving the "ovirt_disk" documentation for disk extend flow

barpavel:
If the disk is referenced by `name` and is attached to a VM, make sure to specify `vm_name` / `vm_id` to prevent extension of another disk that is not attached to the VM.

Bug-Url: https://bugzilla.redhat.com/2016638

### DIFF
--- a/plugins/modules/ovirt_disk.py
+++ b/plugins/modules/ovirt_disk.py
@@ -73,6 +73,8 @@ options:
             - "Size of the disk. Size should be specified using IEC standard units.
                For example 10GiB, 1024MiB, etc."
             - "Size can be only increased, not decreased."
+            - "If the disk is referenced by C(name) and is attached to a VM, make sure to specify C(vm_name)/C(vm_id)
+               to prevent extension of another disk that is not attached to the VM."
         type: str
     interface:
         description:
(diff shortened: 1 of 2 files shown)


plugins/modules/ovirt_disk.py
```diff
@@ -73,6 +73,7 @@
             - "Size of the disk. Size should be specified using IEC standard units.
                For example 10GiB, 1024MiB, etc."
             - "Size can be only increased, not decreased."
+            - "If there are multiple disks with the same name, provide C(vm_name)/C(vm_id) and/or disk C(id) to uniquely identify the intended disk."
```
ahadas, line 76: I think the first part (`If there are multiple disks with the same name`) is a bit ambiguous because it's not clear if it means several disks that are attached to the same VM or let's say when we have a disk that belongs to a single template and there are multiple vms that are provisioned (with thin-provisioning) on top of that template, whether it means we have multiple disks or it's considered to be a single disk..
how about being a bit more directed here - we want to say that if the disk is attached to a vm then it's always recommended to specify the vm name/id, right? in that case, I'd go with something similar to what I suggested on bugzilla: `if the disk is referenced by name and is attached to a VM, make sure to specify vm_name or vm_id to prevent extension of another disk with the same name`

barpavel: No problem, will fix now.
By my phrasing I wanted to cover all possible cases: disks attached to one VM, disk with the same name attached to another VM, not attached to any VM, **several disks with the same name attached to the same VM**,...
And I didn't want a too long phrasing that will also include the case with several disks with the same name attached to the same VM.
I wanted the user to understand from the "tip" that he needs to provide the necessary information to uniquely identify the disk. It could be a `vm_name` / `vm_id` or in some extreme cases even disk `id`.
But, again, no problem, will change now.

barpavel: Done.

mnecas (approved):
LGTM

PR #561, opened 2022-07-12
hosted_engine_setup: fix HE VM networking on recent CentOS Stream

michalskrivanek:
When HostedEngineLocal boots up it applies the cloud init configuration.
For some reason it stopped working recently on dual stack where the VM
should be configured on IPv6 with IPv6 route set up to be reachanble
from host-0. Even though we configure dhcp6 on eth0 in cloud-init config
the route is not configured. Let's keep it as "auto" and let the OS pick
up the right thing, it seems to work ok.


michalskrivanek:
closed in favor of https://github.com/oVirt/ovirt-appliance/pull/72

PR #562, opened 2022-07-17
Improving "ovirt_disk" (mostly documentation)

barpavel:
1. Improving `ovirt_disk` documentation comments (minor rephrasing + typos).
WIP


mwperina:
@barpavel Anything else missing? If not, then please move the PR to review ...

barpavel:
> @barpavel Anything else missing? If not, then please move the PR to review ...

Yes, there were few more things I wanted to add, the PR is not finished.
I will try to finish it this week.

barpavel:
Actually 1 of the reasons this PR started, was that I saw the following strange documentation, see https://docs.ansible.com/ansible/latest/collections/ovirt/ovirt/ovirt_disk_module.html
```
bootable boolean

True if the disk should be bootable. By default when disk is created it isn’t bootable.
Choices:
no
yes
```
I mean pay attention to the text that requires `True` vs. "Choices` that have other enum values....

barpavel:
> Actually one of the reasons this PR started, was that I saw the following strange documentation, see https://docs.ansible.com/ansible/latest/collections/ovirt/ovirt/ovirt_disk_module.html
> 
> ```
> bootable boolean
> 
> True if the disk should be bootable. By default when disk is created it isn’t bootable.
> Choices:
> no
> yes
> ```
> 
> I mean pay attention to the text that requires `True` vs. `Choices` that have other enum values: `yes` /`no`.... I wanted to fix that but I'm not sure should I and how...

@mnecas @mwperina What do you think?

mwperina (approved):
+1

plugins/modules/ovirt_disk.py
```diff
@@ -579,8 +578,8 @@ def download_disk_image(connection, module):
 
 
 def upload_disk_image(connection, module):
-    transfers_service = connection.system_service().image_transfers_service()
-    hosts_service = connection.system_service().hosts_service()
+    connection.system_service().image_transfers_service()
+    connection.system_service().hosts_service()
```
mwperina, line 582: Why do we need to fetch those services when they are not used anywhere?

plugins/modules/ovirt_disk.py
```diff
@@ -65,7 +65,7 @@
                if you want to upload the disk even if the disk with C(id) or C(name) exists,
                then please use C(force) I(true). If you will use C(force) I(false), which
                is default, then the disk image won't be uploaded."
-            - "Note that to upload iso the C(format) should be 'raw'"
+            - "Note that in order to upload iso the C(format) should be 'raw'".
```
mwperina, line 68: I think `.` should be moved ahead of `"`

barpavel: You are right, my bad :)
Fixed.

plugins/modules/ovirt_disk.py
```diff
@@ -552,8 +551,8 @@ def finalize_transfer(connection, module, transfer_id):
 
 
 def download_disk_image(connection, module):
-    transfers_service = connection.system_service().image_transfers_service()
-    hosts_service = connection.system_service().hosts_service()
+    connection.system_service().image_transfers_service()
+    connection.system_service().hosts_service()
```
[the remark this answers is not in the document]
barpavel: Actually I didn't go deep into code to check whether there are some "side effects" of the call.
Many times these methods are more than just a simple getter.
I.e., they open/initialize some service/connection and then return some service ID or handler or something. So I didn't delete the calls, since didn't want to introduce any bugs...

plugins/modules/ovirt_disk.py
barpavel: Same answer as above.
Currently I have a limited availability to start testing this, but if you / @mnecas are sure it's safe, I can remove those calls at all :)

mnecas: If it is not used anywhere than it can be removed. Looks like it is some leftover code from the manual upload/download of disks.

barpavel: Done :)

plugins/modules/ovirt_disk.py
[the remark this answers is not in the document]
barpavel: Done.
Removed.

mwperina (approved):
+1

PR #567, opened 2022-07-28
roles: hosted_engine_setup: restore - remove host also based on name

arachmani:
In restore, remove the host from DB based on the name
in addition to the UUID.
Using multiple hosts with the same name is not allowed anyway.

Bug-Url: https://bugzilla.redhat.com/2003515

### DIFF
--- a/changelogs/fragments/567-hosted_engine_setup-remove-host-based-on-name.yml
+++ b/changelogs/fragments/567-hosted_engine_setup-remove-host-based-on-name.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - hosted_engine_setup - restore - remove host also based on name (https://github.com/oVirt/ovirt-ansible-collection/pull/567).
--- a/roles/hosted_engine_setup/tasks/restore_backup.yml
+++ b/roles/hosted_engine_setup/tasks/restore_backup.yml
@@ -49,7 +49,7 @@
   include_tasks: restore_host_redeploy.yml
   loop:
     - { vds_type: 'vds_name', input: "{{ he_host_name }}" }
-    - { vds_type: 'vds_id', input: "{{ hostvars[he_ansible_host_name]['unique_id_out']['stdout_lines']|first }}" }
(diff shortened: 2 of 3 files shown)


mwperina (approved):
+1

didib:
OK for me, but perhaps better warn the user if there are any inconsistencies - at least if we removed more than one host (e.g. one matching uuid and one matching name).

arachmani:
> OK for me, but perhaps better warn the user if there are any inconsistencies - at least if we removed more than one host (e.g. one matching uuid and one matching name).

Where do you think we need to warn the user?
We already mentioned it in the documentation:
```
If you deploy on a new host, you must assign a unique name to the host. Reusing the name of an existing host included in the backup can cause conflicts in the new environment.
```
 https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.4/html/administration_guide/chap-backups_and_migration#Backing_up_and_Restoring_a_Self-hosted_Engine

didib:
> Where do you think we need to warn the user?

That's a good question. I wrote a bit about it in the bug. Ideally IMO we should prompt and ask the user what to do, e.g. similarly to what we do when 'Add host' fails. I realize this is a large change, so not requiring it right now. I also see it's not easy to emit warnings, https://github.com/ansible/ansible/issues/67260, but perhaps we can still do something, not sure.

arachmani:
Verified locally:
```
2022-08-10 08:57:58,709+0000 DEBUG var changed: host "localhost" var "db_remove_he_host" type "<class 'dict'>" value: "{
    "changed": true,
    "cmd": [
        "/usr/share/ovirt-engine/dbscripts/engine-psql.sh",
        "-c",
        "SELECT deletevds(vds_id) FROM (SELECT vds_id FROM vds WHERE upper(vds_name)=upper('h452u.asrachmani.com')) t"
    ],
    "delta": "0:00:01.201757",
    "end": "2022-08-10 08:57:58.238446",
    "failed": false,
    "msg": "",
    "rc": 0,
    "start": "2022-08-10 08:57:57.036689",
    "stderr": "",
    "stderr_lines": [],
    "stdout": " deletevds \n-----------\n \n(1 row)",
    "stdout_lines": [
        " deletevds ",
        "-----------",
        " ",
        "(1 row)"
    ]
}"
```

PR #569, opened 2022-08-02
hosted_engine_setup: hosted-engine.conf permissions

michalskrivanek:
- hosted_engine_setup: fix HE VM networking on recent CentOS Stream
- align hosted-engine.conf permissions to host-deploy

### DIFF
--- a/changelogs/fragments/569-heconf-permissions.yml
+++ b/changelogs/fragments/569-heconf-permissions.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+    - hosted_engine_setup - fix hosted-engine.conf permissions and ownership (https://github.com/oVirt/ovirt-ansible-collection/pull/569).
--- a/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
+++ b/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
@@ -147,15 +147,22 @@
       path: /var/run/ovirt-hosted-engine-ha
       state: directory
       mode: 0755
-  - name: Copy configuration files to the right location on host
+  - name: Copy vm.conf to the right location on host
     ansible.builtin.copy:
       remote_src: true
-      src: "{{ item.src }}"
-      dest: "{{ item.dest }}"
-      mode: 0644
-    with_items:
-      - {src: "{{ he_local_vm_dir }}/vm.conf", dest: /var/run/ovirt-hosted-engine-ha}
-      - {src: "{{ he_local_vm_dir }}/hosted-engine.conf", dest: /etc/ovirt-hosted-engine/}
+      src: "{{ he_local_vm_dir }}/vm.conf"
+      dest: "/var/run/ovirt-hosted-engine-ha"
+      owner: 'vdsm'
+      group: 'kvm'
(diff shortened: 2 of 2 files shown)


michalskrivanek:
```
[root@ost-he-basic-suite-master-host-0 ~]# ls -l /etc/ovirt-hosted-engine/
total 12
-rw-r--r--. 1 root root  224 Aug  1 06:49 10-appliance.conf
-r--r-----. 1 vdsm kvm  1033 Aug  2 13:35 hosted-engine.conf
-rw-------. 1 root root  103 Aug  2 02:30 virsh_auth.conf
[root@ost-he-basic-suite-master-host-0 ~]# ls -l /var/run/ovirt-hosted-engine-ha
total 4
-rw-r-----. 1 vdsm kvm 1306 Aug  2 13:35 vm.conf
```

mwperina (approved):
+1

arachmani:
Don't we need to change permissions also in this task:
https://github.com/oVirt/ovirt-ansible-collection/blob/509ad0c48a9489bdd811da9331174b694a411b75/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml#L121

How does the first point in the first comment related to this PR -
`hosted_engine_setup: fix HE VM networking on recent CentOS Stream` ?

mwperina:
> Don't we need to change permissions also in this task:
> 
> https://github.com/oVirt/ovirt-ansible-collection/blob/509ad0c48a9489bdd811da9331174b694a411b75/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml#L121
> 
> How does the first point in the first comment related to this PR - `hosted_engine_setup: fix HE VM networking on recent CentOS Stream` ?

Asaf, could you please provide a patch for that?

michalskrivanek:
> Don't we need to change permissions also in this task:
> 
> https://github.com/oVirt/ovirt-ansible-collection/blob/509ad0c48a9489bdd811da9331174b694a411b75/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml#L121
> 
> How does the first point in the first comment related to this PR - `hosted_engine_setup: fix HE VM networking on recent CentOS Stream` ?

is the he_local_vm_dir not being cleared at the end? It has all sorts of temporary files and configurations used during installation, I don't think these permissions are going to solve all of them. We shouldn't leave this behind, and if we indeed don't then it's fine as is

michalskrivanek:
> > Don't we need to change permissions also in this task:
> > https://github.com/oVirt/ovirt-ansible-collection/blob/509ad0c48a9489bdd811da9331174b694a411b75/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml#L121
> > 
> > How does the first point in the first comment related to this PR - `hosted_engine_setup: fix HE VM networking on recent CentOS Stream` ?
> 
> is the he_local_vm_dir not being cleared at the end? It has all sorts of temporary files and configurations used during installation, I don't think these permissions are going to solve all of them. We shouldn't leave this behind, and if we indeed don't then it's fine as is

it seems to me it's cleaned in all cases - failures as well as full and partial execution.

arachmani (approved):
LGTM

PR #573, opened 2022-08-03
hosted_engine_setup: Detect hosted-engine-ha version using /usr/libexec/platform-python

arachmani:
Bug-Url: https://bugzilla.redhat.com/2114769

### DIFF
--- a/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
+++ b/roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
@@ -111,7 +111,7 @@
     changed_when: true
   - name: Detect ovirt-hosted-engine-ha version
     ansible.builtin.command: >-
-      {{ ansible_python.executable }} -c
+      /usr/libexec/platform-python -c
       'from ovirt_hosted_engine_ha.agent import constants as agentconst; print(agentconst.PACKAGE_VERSION)'
     environment: "{{ he_cmd_lang }}"
     register: ha_version_out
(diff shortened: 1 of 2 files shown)


roles/hosted_engine_setup/tasks/create_target_vm/03_hosted_engine_final_tasks.yml
```diff
@@ -110,9 +110,8 @@
     ignore_errors: true
     changed_when: true
   - name: Detect ovirt-hosted-engine-ha version
-    ansible.builtin.command: >-
-      {{ ansible_python.executable }} -c
-      'from ovirt_hosted_engine_ha.agent import constants as agentconst; print(agentconst.PACKAGE_VERSION)'
+    ansible.builtin.shell: >-
+      rpm -qi ovirt-hosted-engine-ha | grep "Version" | awk '{print $3}'
```
kobihk, line 114: maybe you can try this one:
```suggestion
      rpm -q --queryformat '%{VERSION}' ovirt-hosted-engine-ha
```

arachmani: Thanks, CI fails with the following error:
`command-instead-of-module rpm used in place of yum or rpm_key module` 
that comes from ansible-lint, so I'll try with the `yum` module.

mwperina (commented):
Wouldn't it be more correct to use package facts to get a version of an installed package? For example:
https://github.com/oVirt/ovirt-engine/blob/master/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-facts/tasks/host-os.yml#L11
https://github.com/oVirt/ovirt-engine/blob/master/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-facts/tasks/host-os.yml#L31

michalskrivanek:
or just use /usr/libexec/platform-python

arachmani:
> Wouldn't it be more correct to use package facts to get a version of an installed package? For example: https://github.com/oVirt/ovirt-engine/blob/master/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-facts/tasks/host-os.yml#L11 https://github.com/oVirt/ovirt-engine/blob/master/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-facts/tasks/host-os.yml#L31

Yes, you are right, I already tried that and the only con is that the log is larger and contains irrelevant data for this task. I change it to use ` /usr/libexec/platform-python` as @michalskrivanek suggested

mwperina (approved):
+1

PR #580, opened 2022-08-10
cluster_upgrade: skip host upgrades without anything to update

mnecas:
Issue: When doing cluster_upgrade on the host without any updates the host upgrade is not started. This creates a problem because we are waiting for the `HOST_UPGRADE_FAILED` `HOST_UPGRADE_FINISHED` `HOST_UPGRADE_FINISHED_AND_WILL_BE_REBOOTED` events, if the upgrade is not started the events are not created so we are waiting for something that will never be there. By default from UI it should timeout after 60 min.

### DIFF
--- a/plugins/modules/ovirt_host.py
+++ b/plugins/modules/ovirt_host.py
@@ -587,7 +587,7 @@ def main():
             last_event = events_service.list(max=1)[0]
 
             if module.params['check_upgrade']:
-                ret = hosts_module.action(
+                hosts_module.action(
                     action='upgrade_check',
                     action_condition=lambda host: not host.update_available,
                     wait_condition=lambda host: host.update_available or (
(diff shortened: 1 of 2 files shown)


plugins/modules/ovirt_host.py
```diff
@@ -610,38 +610,44 @@ def main():
...
-                            # 166: No available host was found to migrate VM
-                            search='type=65 or type=140 or type=166 or type=17',
-                        ) if host.name in event.description
-                    ]) > 0
-                ),
-                reboot=module.params['reboot_after_upgrade'],
+            # List all update check events and check if we some updates were found
+            no_upgrade = events_service.list(
+                from_=int(last_event.id),
+                search='type=885 and message="*no updates found*" and host.name=%s' % host.name,
```
mwperina, line 616: When there are no updates available, then HostUpgradeCommand failes with specific message:

https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/hostdeploy/UpgradeHostCommand.java#L50
https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/UpgradeHostValidator.java#L45

Wouldn't it be easier to check this result?

mnecas: Was able to find middle ground... now it will skip the wait if the update is not available

mwperina (approved):
+1

plugins/modules/ovirt_host.py
mwperina: Great!

Issue #586, opened 2022-08-30
ovirt_host: reinstall not accepting params


jwahsnakupaku:
##### SUMMARY
When using reinstall in ovirt_host, the following params aren't accepted;
- activate
- reboot_after_installation

##### COMPONENT NAME
ovirt_host - reinstall

##### STEPS TO REPRODUCE

```yaml

        - name: Reinstall Host
          ovirt.ovirt.ovirt_host:
            activate: false
            address: "{{ ansible_hostname }}"
            auth: "{{ ovirt_auth }}"
            cluster: "{{ cluster }}"
            kdump_integration: disabled
            name: "{{ shortname }}"
            state: reinstalled
            public_key: true
            reboot_after_installation: false
```

##### EXPECTED RESULTS
After reinstall, host doesn't reboot or attempt to activate.
Should be able to configure these params when performing a reinstall.
Activate/reboot are configurable within the UI.

##### ACTUAL RESULTS
Playbook errors, as the host is not in a state to activate yet.

#### 
Had a look at ovirt_host.py and looks like the params can be added in and the activate after made conditional.
```
--- a/plugins/modules/ovirt_host.py
+++ b/plugins/modules/ovirt_host.py
@@ -719,8 +719,10 @@ def main():
             )
 
             # Reinstall host:
-            hosts_module.action(
+            ret = hosts_module.action(
                 action='install',
+                activate=module.params['activate'],
+                reboot=module.params.get('reboot_after_installation'),
                 action_condition=lambda h: h.status == hoststate.MAINTENANCE,
                 post_action=hosts_module.post_reinstall,
                 wait_condition=lambda h: h.status == hoststate.MAINTENANCE,
@@ -740,13 +742,13 @@ def main():
                 ) if module.params.get('hosted_engine') is not None else None,
             )
 
-            # Activate host after reinstall:
-            ret = hosts_module.action(
+            if module.params['activate'] and host is not None:
+                ret = hosts_module.action(
                 action='activate',
                 action_condition=lambda h: h.status == hoststate.MAINTENANCE,
                 wait_condition=lambda h: h.status == hoststate.UP,
                 fail_condition=failed_state,
-            )
+                )
```

patersunny:
Nice

mnecas:
Hi @jwahsnakupaku, thanks for the issue and the solution!
Have posted PR with your suggestion.

PR #587, opened 2022-08-30
ovirt_host: reinstalled - honor activate and reboot_after_installation when false

mnecas:
ovirt_host - Honor activate and reboot_after_installation when they are set to false with reinstalled host state
Fixes: https://github.com/oVirt/ovirt-ansible-collection/issues/586


mnecas:
The CI fails because of ansible-lint. Fix in https://github.com/oVirt/ovirt-ansible-collection/pull/588

mwperina (approved):
+1

PR #588, opened 2022-08-30
CI: Fix ansible-lint

mnecas:
The new version of ansible-lint adds a few new checks this PR makes the errors just warnings and we can fix them in the future.


mwperina (approved):
+1

Issue #593, opened 2022-09-27
ovirtvmipsv4 bug when using attr


piramix:
##### SUMMARY
When I use filter ovirtvmipsv4 with attr then I'm getting wrong results (ips are missing)

##### COMPONENT NAME
ovirtmipsv4 filter

##### STEPS TO REPRODUCE

```yaml
  tasks:
    - ovirt.ovirt.ovirt_vm_info:
        pattern: name=nl*
        auth: "{{ ovirt_auth }}"
        follow: ['reported_devices']
      register: vms
    - name: Print VM IP
      debug:
        msg: "{{ vms.ovirt_vms | ovirt.ovirt.ovirtvmipsv4(attr='name') }}"
```

##### EXPECTED RESULTS
TASK [Print VM IP] **********************************************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": {
        "nlx1": [
            "10.46.15.47"
        ],
        "nlx2": [
            "10.46.15.48"
        ],

##### ACTUAL RESULTS

```paste below
TASK [Print VM IP] **********************************************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": [
        "nlx1",
        "nlx2",
        "nlx3",
        "nlx4",
```

Code:

```
    def ovirtvmipsv4(self, ovirt_vms, attr=None, network_ip=None):
        'Return list of IPv4 IPs'
        ips = self._parse_ips(ovirt_vms, lambda version: version == 'v4', attr)
        resp = [ip for ip in ips if self.__address_in_network(ip, network_ip)]
        return resp
```

This code is returning resp instead of ips and if you will use any attr you'll get wrong results
Proper behavior is when I modified the code and removed resp line and use return ips.

```
    def ovirtvmipsv4(self, ovirt_vms, attr=None, network_ip=None):
        'Return list of IPv4 IPs'
        ips = self._parse_ips(ovirt_vms, lambda version: version == 'v4', attr)
        return ips
```

piramix:
You have bug in this merge when you setup network_ip='10.0.0.0/8' and there is a list returned for ex ['10.0.0.1','10.0.0.2']
it return that you cannot use split on a list.

this line:
return dict((k, v) for k, v in ips.items() if self.__address_in_network(v, network_ip))
should look like:
return dict((k, ip) for k, v in ips.items() for ip in v if self.__address_in_network(ip, network_ip))

PR #596, opened 2022-10-06
Fix ovirtvmipsv4 when using attr

mnecas:
Fixes: https://github.com/oVirt/ovirt-ansible-collection/issues/593
The issue was that we were using the filter for the list on dict which returns the list of keys but not values.


mwperina (approved):
+1

Issue #595, opened 2022-10-04
Add filter and test documentation


samccann:
##### SUMMARY

ansible-core 2.14 adds 'side car docs' which gives collection owners the ability to add documentation for previously undocumentable filter and test plugins.

This requires creating yaml files in the same directory as the filter or test plugin code, one yaml file per plugin.

See https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/filter/b64decode.yml for an example.
##### ADDITIONAL INFORMATION

```yaml

```
Related to https://github.com/ansible-community/community-team/issues/65

samccann:
Ideal timeline would be to provide these by the time Ansible 7 releases at the end of November but we realize this information wasn't conveyed to collection owners on a more timely basis.

PR #603, opened 2022-10-12
[WIP] Filters: Add docs

mnecas:
Fixes: https://github.com/oVirt/ovirt-ansible-collection/issues/595

### DIFF
--- a/plugins/filter/json_query.yml
+++ b/plugins/filter/json_query.yml
@@ -1,9 +1,9 @@
 DOCUMENTATION:
   name: json_query
-  author: Martin Necas (@mnecas)
   short_description: Copy of community.general.json_query
   description:
-    - Copy of community.general.json_query used internally in the collection.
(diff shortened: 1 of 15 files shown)


plugins/filter/json_query.yml
```diff
@@ -0,0 +1,20 @@
+DOCUMENTATION:
+  name: json_query
+  author: Martin Necas (@mnecas)
```
mwperina, line 3: Shouldn't there be an original author from community.general.json_query ?

mwperina (changes requested):

plugins/filter/json_query.yml
```diff
@@ -0,0 +1,20 @@
...
+  short_description: Copy of community.general.json_query
+  description:
+    - Copy of community.general.json_query used internally in the collection.
```
mwperina, line 6: Please change to 
```
Copy of community.general.json_query used internally in the collection to ease RPM packaging, so we don't need to package/release/support the whole community.general collection for RHV customers
```

mwperina (approved):
+1

PR #607, opened 2022-10-15
filters: Fix ovirtvmipsv4 with attribute and network

mnecas:
Fixes: https://github.com/oVirt/ovirt-ansible-collection/issues/593#issuecomment-1279687954
Issue: When filtering the vms with atr and network the filter failed because we were checking list of ips insead of single ip


mwperina (approved):
+1

PR #618, opened 2022-11-10
Fixing create VM from template with non admin user

emesika:
This patch fixes creating a VM from a template where the template resides in the Default DC.
REST API has a problem of not getting the href for the DC when dealing with Default cluster.
In order to fix that a special treatment was added to get the right Default DC object.
Solution will not work if the Default DC/Cluster name was changed from "Default" to something else. ( we don't have any indication apart of the name for Default DC/Cluster)

Signed-Off-By: <Eli Mesika: emesika@redhat.com>
Bug-Url: https://bugzilla.redhat.com/2078946

### DIFF
--- a/plugins/modules/ovirt_vm.py
+++ b/plugins/modules/ovirt_vm.py
@@ -1414,8 +1414,31 @@ class VmsModule(BaseModule):
         if self._is_new:
             if self.param('template'):
                 clusters_service = self._connection.system_service().clusters_service()
-                cluster = search_by_name(clusters_service, self.param('cluster'))
-                data_center = self._connection.follow_link(cluster.data_center)
+                data_centers_service = self._connection.system_service().data_centers_service()
+                clusters = clusters_service.list(
+                )
+                cluster = [
+                        c for c in clusters
+                        if c.name == self.param('cluster')
+                ]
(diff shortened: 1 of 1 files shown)


plugins/modules/ovirt_vm.py
```diff
@@ -1414,8 +1414,24 @@ def __get_template_with_version(self):
         if self._is_new:
             if self.param('template'):
                 clusters_service = self._connection.system_service().clusters_service()
+                data_centers_service = self._connection.system_service().data_centers_service()
                 cluster = search_by_name(clusters_service, self.param('cluster'))
-                data_center = self._connection.follow_link(cluster.data_center)
+                # For default DC/Cluster we use the hardcoded name 'Default' even that name
+                # can be modified, we have no other indication to know that.
+                # The API does not return a href for the data_center of the default cluster
```
mwperina, line 1421: Why API behaves differently for the default DC? What's the reason to not return the href to the default cluster same as for the other clusters?

@oliel

mwperina (changes requested):

mwperina:
In favor of https://github.com/oVirt/ovirt-engine/pull/766

PR #621, opened 2022-11-14
Wait for host to be in result state

dupondje:
Bug introduced in #580.
The update_available is false just after the upgrade, and HOST_UPGRADE_FINISHED_AND_WILL_BE_REBOOTED is logged. But the host is not yet rebooted, so we still need to wait for the status to be UP or MAINTENANCE.


plugins/modules/ovirt_host.py
```diff
@@ -613,7 +613,7 @@ def main():
             ret = hosts_module.action(
                 action='upgrade',
                 action_condition=lambda h: h.update_available,
-                wait_condition=lambda h: not h.update_available or h.status == result_state and (
+                wait_condition=lambda h: not h.update_available and h.status == result_state and (
```
mnecas, line 616: Is the update_availible False even when the upgrade fails? 
If it would be still available the module would still wait and timeout.

dupondje: I should check.
But even then, the status will never be result_state (but FailedInstall), so it will always wait and timeout no?

dupondje: Better logic in the next commit. That should cover the cases like before.

plugins/modules/ovirt_host.py
```diff
@@ -625,7 +627,7 @@ def main():
                             search='type=842 or type=841 or type=888',
                         ) if host.name in event.description
                     ]) > 0
-                ),
+                ) or not updates_available),
```
mnecas, line 630: The `host` variable does not update over time compared to the `h` in lambda. So when the update would not available the wait_condition will be always true and the timeout.

mnecas (changes requested):

plugins/modules/ovirt_host.py
dupondje: But that is just what we want.
- If there was an update available before the upgrade process started, then you want to wait for event 842/841/888 to make sure host has been rebooted AND is again in the result_state (MAINTENANCE or UP).
- If there was no update available, then we don't have to wait neither.

mnecas (approved):

plugins/modules/ovirt_host.py
mwperina: You are right, now we understand the fix, thanks

mwperina (approved):
+1

PR #637, opened 2022-11-28
cluster_upgrade: Fix the engine_correlation_id location

mnecas:
Issue:
When there is `random` filter in the ansible defaults, and the value is not specified, it does not generate one random number per run but per each use of the unset variable.
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2142601


mwperina (approved):
+1

PR #647, opened 2022-12-07
Add default_route usage to the ovirt_network module

ilush:
Currently, it is possible to configure default_route usage through web-UI or API, but not through ansible.


mwperina (changes requested):
Could you please add changelog fragment documenting this change?

Other than that it looks good to me.

mwperina (approved):

mnecas (changes requested):
Please add the `default_route` to the `DOCUMENTATION`

mnecas (approved):
LGTM

PR #653, opened 2022-12-12
info modules: use @NAMESPACE@.@NAME@ instead of ovirt.ovirt for deprecation warning

mnecas:
info modules: use @NAMESPACE@.@NAME@ instead of ovirt.ovirt for deprecation warning

Signed-off-by: Martin Nečas <necas.marty@gmail.com>


mwperina (approved):
+1

PR #657, opened 2022-12-21
Relax bindep package specifiers to support EL9

shanemcd:
Hello. I ran into issues when trying to build [awx-ee](https://github.com/ansible/awx-ee) on top of stream9 images. These changes were necessary in order for ansible-builder to properly auto-detect the build dependencies on EL9.


AlanCoding:
:+1: this has always been a pain point, and it's hard to imagine that this would do harm.

bindep.txt
```diff
@@ -1,6 +1,7 @@
-gcc [compile platform:centos-8 platform:rhel-8]
-libcurl-devel [compile platform:centos-8 platform:rhel-8]
-libxml2-devel [compile platform:centos-8 platform:rhel-8]
-openssl-devel [compile platform:centos-8 platform:rhel-8]
+gcc [compile platform:rpm]
+libcurl-devel [compile platform:rpm]
+libxml2-devel [compile platform:rpm]
+openssl-devel [compile platform:rpm]
 python38-devel [compile platform:centos-8 platform:rhel-8]
```
mwperina, line 5: Could you please change  above line to below?
```
python39-devel [compile platform:centos-8 platform:rhel-8]
```

I forgot to do that as a part of https://github.com/oVirt/ovirt-ansible-collection/pull/632

mwperina (changes requested):
Generally looks great, one minor change and it can be merged

bindep.txt
shanemcd: Oops, this almost fell off my radar. Went ahead and made that change @mwperina. This should be good to go now.

mwperina: Thanks

mwperina (approved):
+1

shanemcd:
@mwperina Thanks for the approval. I just noticed the CI checks are failing but don't seem to be related to anything I've changed here. Do you have an ETA of when this might get cut into a release on Galaxy? I _would_ point to the git repo once this merges but the way galaxy.yml is templated makes this impossible.

mwperina:
> @mwperina Thanks for the approval. I just noticed the CI checks are failing but don't seem to be related to anything I've changed here. Do you have an ETA of when this might get cut into a release on Galaxy? I _would_ point to the git repo once this merges but the way galaxy.yml is templated makes this impossible.

We don't have an exact schedule, but most probably we will have a new stable release around end of January.

PR #661, opened 2023-01-16
roles: hosted_engine_setup: Delete also the bridge

didib:
Recently CI CentOS Stream 9 fails on the following `virsh net-start default` saying that virbr0 already exists. Not sure why it's not deleted by `virsh net-destroy default` - perhaps that's a change in libvirt. Let's remove the bridge explicitly.

Signed-off-by: Yedidyah Bar David <didi@redhat.com>

### DIFF
--- a/changelogs/fragments/661-delete-also-the-bridge.yml
+++ b/changelogs/fragments/661-delete-also-the-bridge.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - he-setup - recently `virsh net-destroy default` doesn't delete the `virbr0`, so we need to delete it expicitly (https://github.com/oVirt/ovirt-ansible-collection/pull/661).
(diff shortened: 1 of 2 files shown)


mwperina (approved):
+1

arachmani (approved):
LGTM

michalskrivanek:
I checked the run - unsure if it is related to this PR. The appliance VM doesnt' seem to be booting, there's this in the console:

You are in emergency mode. After logging in, type "journalctl -xb" to view
system logs, "systemctl reboot" to reboot, "systemctl default" or "exit"
to boot into default mode.

Cannot open access to console, the root account is locked.
See sulogin(8) man page for more details.

Press Enter to continue.

Reloading system manager configuration
[  446.084682] systemd-rc-local-generator[1692]: /etc/rc.d/rc.local is not marked executable, skipping.
Starting default.target

michalskrivanek:
eh, gh webhooks misbehaving again:/

michalskrivanek:
the associated builds seem to have a problem (again, i think it didnt' really trigger them right), can you try to push another change, even a dummy one?

PR #662, opened 2023-01-17
Test, do not merge: Debug appliance virt install failure

didib:


didib:
Not needed anymore, and was not enough. See https://github.com/oVirt/ovirt-appliance/pull/108#issuecomment-1386608964 for what eventually worked for me locally, if you need to investigate similar issues.

Issue #659, opened 2023-01-11
OBSOLETE: method inspect.getargspec does no longer exist in python 3.11! Drop in replacement should be inspect.getfullargspec.


6ramr:
##### SUMMARY
The method inspect.getargspec is now obsolete and is no longer featured in python 3.11. Documentation suggests inspect.getfullargspec, which should work as drop in replacement in most cases.

https://docs.python.org/3.11/library/inspect.html
https://docs.python.org/3.10/library/inspect.html
##### COMPONENT NAME
Seen in  
_plugins/module_utils/ovirt.py_  
but might exist elsewhere
##### STEPS TO REPRODUCE

```
- name: Create Network
  hosts: localhost
  tasks:
     - name: ensuring customer network exists
      ovirt.ovirt.ovirt_network:
        auth:
          insecure: yes
        data_center: myDatacenter
        name: mynetwork
        vm_network: true 
        external_provider: ovirt-provider-ovn
        description: this test is
```

##### EXPECTED RESULTS
Creates network

##### ACTUAL RESULTS

```
The full traceback is:
Traceback (most recent call last):
  File "/tmp/ansible_ovirt.ovirt.ovirt_network_payload_gh3hf6ez/ansible_ovirt.ovirt.ovirt_network_payload.zip/ansible_collections/ovirt/ovirt/plugins/modules/ovirt_network.py", line 341, in main
  File "/tmp/ansible_ovirt.ovirt.ovirt_network_payload_gh3hf6ez/ansible_ovirt.ovirt.ovirt_network_payload.zip/ansible_collections/ovirt/ovirt/plugins/module_utils/ovirt.py", line 351, in get_id_by_name
    entity = search_by_name(service, name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/ansible_ovirt.ovirt.ovirt_network_payload_gh3hf6ez/ansible_ovirt.ovirt.ovirt_network_payload.zip/ansible_collections/ovirt/ovirt/plugins/module_utils/ovirt.py", line 311, in search_by_name
    if 'search' in inspect.getargspec(service.list)[0]:
                   ^^^^^^^^^^^^^^^^^^
AttributeError: module 'inspect' has no attribute 'getargspec'
fatal: [localhost]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "clusters": null,
            "comment": null,
            "data_center": "myDatacenter",
[...]

PR #663, opened 2023-01-18
Replace getargspec with getfullargspec

mnecas:
Fixes: https://github.com/oVirt/ovirt-ansible-collection/issues/659
The `getargspec` was supported in python 2 and removed in python 3.11.
The alternative is `getfullargspec`, which is in all python 3 versions.


mwperina (approved):
+1

PR #664, opened 2023-01-18
Debug appliance virt install failure

didib:


michalskrivanek:
I think it's a good idea to have more debugging always, LGTM

mwperina (approved):
+1

paralk:
Using `virsh console HostedEngineLocal --force" releases the console held by ansible and the tasks then proceed. However would like to avoid this manual intervention somehow...

paralk:
Adding to the remote host the variable `ansible_ssh_use_tty=false` gets past the "Create Local VM" task with the error message 
"error: Cannot run interactive console without a controlling TTY" i.e. the same as when running `hosted-engine`

didib:
> @didib @mnecas @mwperina When deploying hosted_engine_setup via an Ansible playbook onto a remote host from local, the "Create Local VM" task that invokes `virt-install` hangs as it appears that Ansible is connecting to the guest VMs serial console and not disconnecting, probably due to this change.

Perhaps

> 
> When hosted engine setup is run on the host itself using `hosted-engine`, the task does not hang. Instead the error message in the logs for virt-install is "error: Cannot run interactive console without a controlling TTY", however the rest of the setup then proceeds fine.
> 
> Are remote deployments from local working for you?

I don't recall ever trying this, personally. I think others did, sporadically, over time, but IIUC it was never considered a supported feature/RFE, and was never part of CI. I vaguely recall that others in Red Hat did this routinely when RHV was in full support, but have no details. So it should probably not be very hard (or very broken).

> Is that something that should not be done at all?

If for some reason this is important for you, by all means go ahead and try, and push patches if needed. Otherwise, I wouldn't bother and just run it on the host.

> 
> Is there a way to have the serial console disconnect without manual intervention and let the rest of the Ansible proceed?

(No idea, but see below)

> Using `virsh console HostedEngineLocal --force` on the host releases the console held by ansible from local, and the rest of the tasks then proceed. However would like to avoid this manual intervention somehow...

> Adding to the remote host the variable `ansible_ssh_use_tty=false` gets past the "Create Local VM" task without manual intervention - with the error message "error: Cannot run interactive console without a controlling TTY" i.e. the same as when running `hosted-engine`
> 
> The rest of the playbook runs ok to the end

Either (or both? not sure) of these sound like a good fix, if you can find the place to add them. I suggest to create a new issue, to properly document this, and a PR, if you come up with a fix. Thanks!

paras-tsl:
thanks for the response

ansible_ssh_use_tty=false on its own fix this issue

 I'll create a new issue, I think as its for remote deployments from a local machine it perhaps simply needs to be documented in the readme

paralk:
Thanks for the response

`ansible_ssh_use_tty=false` on its own fixes this issue

Will raise a new issue and PR

PR #676, opened 2023-02-08
Python 3.11 has dropped support for getargspec

snecklifter:
This is a temporary patch in case we still want to support older versions

### DIFF
--- a/plugins/module_utils/ovirt.py
+++ b/plugins/module_utils/ovirt.py
@@ -25,6 +25,10 @@ import inspect
 import os
 import time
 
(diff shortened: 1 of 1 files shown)


mnecas (commented):
Do we still use the `inspect.getargspec` I replaced it with `getfullargspec` in
https://github.com/oVirt/ovirt-ansible-collection/pull/663

snecklifter:
> Do we still use the `inspect.getargspec` I replaced it with `getfullargspec` in #663

You did and I missed that, its just not made it into a release yet. Thanks!

PR #688, opened 2023-03-01
spec: Add python3.11 requirements

mnecas:
The ansible-core 2.14 uses python3.11 on Centos 9 stream instead of the default system python3.9. We need to build the rpm to use the collection with the ansible-core 2.14, which forces the python library to python3.11 so the ansible can use the collection.


mwperina (approved):
+1

mwperina (commented):
Please add more information into commit description about the Python 3.11 requirements

Issue #695, opened 2023-03-22
oVirt hosted engine installation failure because of missing netaddr for python 3.11 in ovirt-node image


daliborfilus:
##### SUMMARY

Don't know if this is the correct place for this issue, plase, redirect me if it isn't.

##### COMPONENT NAME

oVirt hosted engine installation procedure.

##### STEPS TO REPRODUCE

As a new user... go to the download page.

1) Download ovirt-node 4.5.4, 4.5.3, 4.5.2 ISO (el8).
2) Install the ovirt-node image, reboot.
3) Run `hosted-engine --deploy --4`, fill-in the form.
4) After ~twenty minutes, watch it hang on "Wait for the host to be up" message... for another tens of minutes. Then it crashes and rolls back the installation.
5) Optional steps: Pull your hair out trying to find something relevant on this issue via google. Find nothing meaningful. Retry the whole installation multiple times with different versions of the image and network configuration. (I thought the issue was that the engine VM can't connect to the host, which led me to trying to fix non-existing issues with my DNS and other network stuff.) I reinstalled the node image 6+ times total. I re-run the hosted-engine deploy commands 10+ times. Two days of life are gone.
6) Finally discover that you can go to the failed VM directly (it's still present and active) and hunt for logs there. 
7) Discover `/var/log/ovirt-engine/engine.log`. See this inside:

```
2023-03-22 15:55:34,168+01 ERROR [org.ovirt.engine.core.bll.hostdeploy.InstallVdsInternalCommand] (EE-ManagedThreadFactory-engine-Thread-1) [9c814912-05ec-49eb-8e4f-912347e59f0d] Host installation failed for host 'e2fc0443-9a67-4c11-a11e-791903212bc2', 'ovirtnode1.b-one.cz': Task Install ovs failed to execute. Please check logs for more details: /var/log/ovirt-engine/host-deploy/ovirt-host-deploy-ansible-20230322155412-ovirtnode1.b-one.cz-9c814912-05ec-49eb-8e4f-912347e59f0d.log
```

Go to that log and see this inside:

```
[...]

PR #696, opened 2023-03-23
Drop usage of ipaddr filters and remove dependency on python-netaddr

michalskrivanek:


michalskrivanek:
nto sure if it fixes #695 but it's a good thing in any case...

michalskrivanek:
seems to work ok, it now broke on the next dependency, jmespath.

mnecas (approved):
Bit surprised that ip rule can accept IP/MASK instead of IP/CIDR. But looks good

michalskrivanek:
tested on el9 with ipv4-only and dual stack

michalskrivanek:
> Bit surprised that ip rule can accept IP/MASK instead of IP/CIDR. But looks good

ah, so that's what it does!:-D dunno, but if the sacred OST says green then why would I even check

michalskrivanek:
That’s one small step for man, one giant leap for mankind!

mnecas:
> > > Bit surprised that ip rule can accept IP/MASK instead of IP/CIDR. But looks good
> > 
> > 
> > ah, so that's what it does!:-D dunno, but if the sacred OST says green then why would I even check
> 
> reading documentation despite them trying to document the syntax using formal grammar it doesn't really seem to be defined what format it accepts. In other places ip's PREFIX states it's CIDR. Yet, it indeed works just fine, even on RHEL 8.6:
> 
> ```
> # ip rule add from 192.168.8.3/255.255.255.0 priority 101 table main
> # ip rule
> 101:    from 192.168.8.3/24 lookup main
> ```
> 
> eh....so let's proceed...:)

Yeah I did same test. Also could not find which format they require/support

PR #700, opened 2023-03-30
ovirt_disk: Fix issue in detaching the direct LUN

nijinashok:
The direct LUN doesn't have a disk profile and follow_link(self._connection, entity.disk_profile) will be None. So detaching the LUN will fail with error "'NoneType' object has no attribute 'name'".

RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=2183112


mwperina (approved):
+1

mnecas (approved):

PR #701, opened 2023-04-04
we need python3-jmespath for ansible runs with platform python

michalskrivanek:
hosted-engine-setup runs with platform-python and requires jmespath


michalskrivanek:
tested manually since default repos need to be added

PR #702, opened 2023-04-17
drop remaining netaddr

michalskrivanek:
commit fd20e99066c34006f392409b96150b29d0968470 dropped python-netaddr
dependency, remove remaining usage of filters requiring it

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -52,7 +52,7 @@ jobs:
             "ansible-lint>=6.0.0,<7.0.0" \
             cryptography \
             "pyyaml>=5.4.1"
-          ansible-lint roles/* -x 204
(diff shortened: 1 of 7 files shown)


mwperina (approved):
+1

michalskrivanek:
fix works. however test_upload_cirros_image is screwed up both on el8 and el9 but I believe that's a different issue

.github/workflows/build.yaml
```diff
@@ -49,7 +52,7 @@ jobs:
             "ansible-lint>=6.0.0,<7.0.0" \
             cryptography \
             "pyyaml>=5.4.1"
-          ansible-lint roles/* -x 204
+          ansible-lint -x 'key-order[task],no-free-form,schema[vars],yaml[octal-values],risky-shell-pipe,fqcn[keyword]' roles/*roles/*
```
mnecas, line 55: Think you have typo in `roles/*roles/*` should be `roles/*`

mnecas (changes requested):

PR #705, opened 2023-04-18
Support boot disk renaming and resizing.

mmartinv:
Renames the boot disk when the `boot_disk_name` variable is defined
within the profile or the VM properties. (The latter has precedence
over the former). The VM disk must have the bootable flag set.

By default, if `boot_disk_name` is defined, the VM name will be
prepended to the boot disk name so the final name would be:

`{{ vm_name }}_{{ boot_disk_name }}`

This behavior can be disabled by setting the variable
`boot_disk_use_vm_name_prefix` to `false`. In this case the
final name of the boot disk will be the value of `boot_disk_name`.

The boot disk can also be resized by defining the `boot_disk_size`
variable.

The value can contain a suffix that complies to the IEC 60027-2
standard (for example 10GiB, 1024MiB)

Signed-off-by: Miguel Martín <mmartinv@redhat.com>

### DIFF
--- a/changelogs/fragments/705-support-boot-disk-resizing-renaming.yml
+++ b/changelogs/fragments/705-support-boot-disk-resizing-renaming.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+      - vm_infra - Support boot disk renaming and resizing. (https://github.com/oVirt/ovirt-ansible-collection/pull/705)
--- a/roles/vm_infra/tasks/rename_resize_boot_disk.yml
+++ b/roles/vm_infra/tasks/rename_resize_boot_disk.yml
@@ -1,0 +1,59 @@
+---
+- name: "Get the '{{ current_vm.name }}' VM info"
+  ovirt_vm_info:
+    auth: "{{ ovirt_auth }}"
+    pattern: "name={{ current_vm.name }}"
+    fetch_nested: true
+    nested_attributes:
+      - "bootable"
+    all_content: true
+  register: current_vm_info
+
+- name: "Get the '{{ current_vm.name }}' VM boot disk ID"
+  ansible.builtin.set_fact:
+    current_vm_boot_disk_id: "{{ item.id }}"
+  loop: "{{ current_vm_info.ovirt_vms[0].disk_attachments }} "
+  when: item.bootable is defined and item.bootable
+
+- name: "Get the '{{ current_vm.name }}' VM boot disk info"
+  ovirt_disk_info:
+    auth: "{{ ovirt_auth }}"
+    pattern: "id={{ current_vm_boot_disk_id }}"
+    fetch_nested: true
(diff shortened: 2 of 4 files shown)


changelogs/fragments/705-support-boot-disk-resizing-renaming.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+      - vm_infra - Support boot disk renaming and resizing. (https://github.com/oVirt/ovirt-ansible-collection/pull/705)
```
mwperina, line 3: Adding new parameters seems to me like a RFE, so it should be a major change. Am I right @mnecas ?

mwperina (commented):
Generally looks good, even though I don't see the use case.

But we need to agree if i's major or minor change from the point of the release version number

mmartinv:
> Generally looks good, even though I don't see the use case.

Renaming the disks is useful because, right now, the template disk name is maintained so if a lot of VMs are created from a template there will be a lot of disks with the same name.

Resizing the boot disks allows the VMs based on a cloud-init template to have an arbitrary root disk size (The ovirt disk is resized by the role and cloud-init resizes the partition and the filesystem size in the first boot.

> But we need to agree if i's major or minor change from the point of the release version number

I personally don't mind, let's see what @mnecas has to say.

changelogs/fragments/705-support-boot-disk-resizing-renaming.yml
mmartinv: @mnecas

roles/vm_infra/tasks/rename_resize_boot_disk.yml
```diff
@@ -0,0 +1,59 @@
...
+    auth: "{{ ovirt_auth }}"
+    pattern: "id={{ current_vm_boot_disk_id }}"
+    fetch_nested: true
+  register: current_vm_boot_disk_info
+  when: current_vm_boot_disk_id is defined
+
+- name: "Get the Storage Domain info from the '{{ current_vm.name }}' VM boot disk"
+  ovirt_storage_domain_info:
+    auth: "{{ ovirt_auth }}"
+    id: "{{ current_vm_boot_disk_info.ovirt_disks[0].storage_domains[0].id }}"
```
mnecas, line 29: Not sure if the module accepts the `id` you might need to use the `pattern` with the id filter. And it returns a list of storage domains.

mmartinv: You are absolutely right and that's why I was using `ovirt_storage_domain` in the first place, it's not possible to retrieve the information of a storage domain with `ovirt_storage_domain_info` if you only have the storage domain id :(

mnecas: resolved offline

PR #707, opened 2023-04-25
ovirt_disk: Fix issue in detaching the direct LUN

michalskrivanek:
The direct LUN doesn't have a disk profile and follow_link(self._connection, entity.disk_profile
) will be None. So detaching the LUN will fail with error "'NoneType' object has no attribute 'name'".

Signed-off-by: Nijin Ashok <nashok@redhat.com>


michalskrivanek:
backport

michalskrivanek:
@mnecas can you please check the linting, whether it's a missing exclude or if it's ok to disable completely

mnecas (approved):

PR #710, opened 2023-07-03
ovirt_quota: Convert storage size to integer

### LINKED ISSUE #709: Can't set storage size limit for quota when using Ansible variables
SUMMARY:

The current implementation of this module seems not to be able to set storage size limit quota while using Ansible variables like e.g.  "{{ storage_size }}"

It seems that "limit=storage.get('size')" must be "limit=int(storage.get('size'))" instead.

https://github.com/oVirt/ovirt-ansible-collection/blob/9befbcf12fd942a961daca9ab3aaa178a59cd76d/plugins/modules/ovirt_quota.py#L312C31-L312C50

hambuergaer:
Converting storage size to integer.
Fixes: #709

### DIFF
--- a/changelogs/fragments/701-fix-jmespath.yml
+++ b/changelogs/fragments/701-fix-jmespath.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+      - HE - add back dependency on python3-jmespath (https://github.com/oVirt/ovirt-ansible-collection/pull/701)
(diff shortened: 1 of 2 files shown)


hambuergaer:
Hi @mnecas, I saw the build pipeline failed due to missing changelog. So I tried to follow this guide here: https://github.com/oVirt/ovirt-ansible-collection/wiki/Release. The problem is that in the build.sh version 3.1.3 is already included but even if version 3.1.2 is current there are still some changelog entries for version 3.1.1 included... 

How should I proceed? Will you fix the changelog?

PR #720, opened 2023-08-24
ovirt_disk: Fix no_log pass warning

Kloppi313:
Set pass_discard to default false in order to fix
`[WARNING]: Module did not set no_log for pass_discard`


Kloppi313:
Closing as it does not do the trick on an other test task.

mnecas:
Yeah, this is by design in the ansible, and we just match the regex for warning.

https://github.com/ansible/ansible/blob/116948cd1468b8a3e34af8e3671f4089e1f7584c/lib/ansible/module_utils/basic.py#L198

PR #723, opened 2023-09-22
ovirt_role: Fix administrative option when set to False

mnecas:
When user set administrative to false the option was not set and API failed with 400.


mwperina (approved):
+1

PR #733, opened 2023-11-27
image_template: enable image template version name

mmartinv:
- image_template: allow to setup template version name
- image_template: allow to setup template OS and description

### DIFF
--- a/roles/image_template/README.md
+++ b/roles/image_template/README.md
@@ -49,6 +49,7 @@ Role Variables
 | glance_image_provider        | UNDEF (mandatory if qcow_url is not used)            | Name of the glance image provider.                    |
 | glance_image            | UNDEF (mandatory if qcow_url is not used)               | This parameter specifies the name of disk in glance provider to be imported as template. |
 | template_prerequisites_tasks | UNDEF | Works only with qcow image. Specify a path to Ansible tasks file, which should be executed on virtual machine before creating a template from it. Note that qcow image must contain guest agent which reports IP address. |
+| template_version_name      | "base version" | Version name of the template. Allows to create different versions of a template.  |
 
 The `template_disks` List of dictionaries can contain following attributes:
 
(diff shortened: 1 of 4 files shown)


mwperina (approved):
LGTM

roles/image_template/tasks/qcow2_image.yml
```diff
@@ -83,7 +83,8 @@
     - name: Check if template already exists
       ovirt_template_info:
         auth: "{{ ovirt_auth }}"
-        pattern: "name={{ template_name }} and datacenter={{ dc_info.ovirt_datacenters[0].name }}"
+        pattern: "name={{ template_name }} and datacenter={{ dc_info.ovirt_datacenters[0].name }}
+          and version_name={{ template_version_name | default('\"base version\"') }}"
```
mnecas, line 87: Please move the default value to [default yaml](https://github.com/oVirt/ovirt-ansible-collection/blob/master/roles/image_template/defaults/main.yml) or remove the default value as mentioned in previous comment, in which case you would need to check if the `template_version_name` is specified using the `{% if template_version_name is defined %} and version_name={{ template_version_name }} {% endif %}`

roles/image_template/README.md
```diff
@@ -49,6 +49,9 @@ Role Variables
 | glance_image_provider        | UNDEF (mandatory if qcow_url is not used)            | Name of the glance image provider.                    |
 | glance_image            | UNDEF (mandatory if qcow_url is not used)               | This parameter specifies the name of disk in glance provider to be imported as template. |
 | template_prerequisites_tasks | UNDEF | Works only with qcow image. Specify a path to Ansible tasks file, which should be executed on virtual machine before creating a template from it. Note that qcow image must contain guest agent which reports IP address. |
+| template_version_name      | "base version" | Version name of the template. Allows to create different versions of a template.  |
```
mnecas, line 52: What if someone runs this role right after upgrading without specifying the `template_version_name` wouldn't it create a new template or add a name to the existing one? 
I am worried that this could cause unexpected behaviour. 
If it creates a new template I would keep the variable `UNDEF`.

mmartinv: > What if someone runs this role right after upgrading without specifying the `template_version_name` wouldn't it create a new template or add a name to the existing one?

None: if the template already exists, nothing will be done, and if it doesn't, a new one will be created with the default `template_version_name` which is `"base version"`.

>I am worried that this could cause unexpected behaviour. If it creates a new template I would keep the variable `UNDEF`.

All the templates created without specifiying the version name have `"base version"` as the default value. This is by default in the sense that the API will put that name if you don't specify it. So the behavior in case `template_version_name` is not specified should be equivalent to the current behavior.

mnecas (approved):
LGTM

mnecas:
Posted https://github.com/oVirt/ovirt-ansible-collection/pull/735 to fix the ansible-test ci

mwperina (approved):
LGTM

PR #736, opened 2024-01-11
Not possible to attach LUN to running VM through ansible

saksham-oracle:
Issue: Ansible example shows that lun can be attached to vm. The example shows to add lun with lun_id and storage_type attribute. The ansibles scripts fails to execute as it fails to attach disk because it failed to find the lun and associated disk info. The issue was the lun directory structure has lun_id attribute and the lookup code was using just id instead of lun_id.

https://docs.ansible.com/ansible/latest/collections/ovirt/ovirt/ovirt_disk_module.html#parameter-logical_unit/lun_id

Fix: Fix the attribute name that used to get the disk info.

Signed-off-by: Saksham Srivastava <saksham.sa.srivastava@oracle.com>

### DIFF
--- a/plugins/modules/ovirt_disk.py
+++ b/plugins/modules/ovirt_disk.py
@@ -843,7 +843,7 @@ def main():
     host = module.params['host']
     # Fail when host is specified with the LUN id. LUN id is needed to identify
     # an existing disk if already available in the environment.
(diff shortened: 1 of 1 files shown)


mnecas:
Hi @saksham-oracle thank you for the MR, checked out the other places and rather would align the documentation with the code. Posted the fix

Issue #755, opened 2025-01-04
Unmaintained collection: Removal from the Ansible Community Package


mariolenz:
It looks like this collection is effectively unmaintained. According to the current [Ansible Community Package Collections Removal Process](https://docs.ansible.com/ansible/devel/community/collection_contributors/collection_package_removal.html#identifying-and-removing-an-unmaintained-collection-that-has-not-been-deprecated-by-its-maintainers), we will consider removing it in a future version of the Ansible Community Package. Please see [Unmaintained collection: ovirt.ovirt](https://forum.ansible.com/t/39455) for more information.

At least one month after this announcement appears here and on [Bullhorn](https://forum.ansible.com/c/news/bullhorn/17), the Ansible Community Steering Committee will vote on whether this collection is considered unmaintained and will be removed, or whether it will be kept. If it will be removed, this will happen earliest in Ansible 13. Please note that people can still manually install the collection with `ansible-galaxy collection install ovirt.ovirt` even when it has been removed from the Ansible Community Package.

mariolenz:
@mnecas you've done the last release and @shubhaOracle you've done the last commit as far as I can see. So maybe you might be interested in this.

dupondje:
Hi @mariolenz 

Thanks for the notifications here.
You are correct that the ansible collection (and big parts of the rest of oVirt) is a bit unmaintained.

But, we will contribute more and more to oVirt, and will also keep this ansible collection maintained.
I think we just need a bit more time to get everything up and running, pipelines on our end etc.

So yes it's a bit unmaintained atm, but it will change. Just need some weeks/months to get our workforce in place!
I would therefor like to keep oVirt in the Ansible collection.

Is there anything urgent that needs to be picked up to avoid the removal? Then I can have a look to resolve those bits at least.

mariolenz:
@dupondje I don't see anything urgent ATM. Anyway, the fact that you've answered this so fast is a sign that we shouldn't remove the package in the near future. I guess I'll keep this on my list of _possibly_ unmaintained collections, but will not proceed with the removal process for now.

I'll have another look at the collection in some weeks/months, though ;-)

mariolenz:
@dupondje
> You are correct that the ansible collection (and big parts of the rest of oVirt) is a bit unmaintained.
> 
> But, we will contribute more and more to oVirt, and will also keep this ansible collection maintained. I think we just need a bit more time to get everything up and running, pipelines on our end etc.

[...]

PR #762, opened 2025-04-17
Maintenance tasks

peter-boden:
- No longer support older ansible versions that are unsupported in RHEL8/9 (and derivatives)
- Use antsibull-nox to run testing and linting (also in ci)
- Run sanity tests for all supported ansible versions, cleanup sanity ignore files, in accordance with ansible collection requirements
- Remove rhv
- Add a badge to show build state on readme
- Started cleanup of linter warnings

Fixes: #755

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -14,9 +14,7 @@ tags:
   - cloud
   - virtualization
 dependencies:
-  "ansible.netcommon": "<3.0.0"
   "ansible.posix": "*"
-  "ansible.utils": "<2.6.0"
 readme: README.md
 repository: https://github.com/ovirt/ovirt-ansible-collection
 issues: https://github.com/ovirt/ovirt-ansible-collection/issues
--- a/meta/requirements.yml
+++ b/meta/requirements.yml
@@ -1,7 +1,3 @@
 ---
 collections:
-  - name: ansible.netcommon
-    version: <3.0.0
   - name: ansible.posix
-  - name: ansible.utils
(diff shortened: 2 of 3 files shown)


galaxy.yml
```diff
@@ -1,17 +1,22 @@
...
-    - virtualization
+  - collection
+  - infrastructure
+  - linux
+  - cloud
+  - virtualization
+dependencies:
+  "ansible.netcommon": "<3.0.0"
+  "ansible.posix": "*"
+  "ansible.utils": "<2.6.0"
```
mariolenz, line 19: Does this collection really depend on `ansible.netcommon<3.0.0` and `ansible.utils<2.6.0`? Those versions are some three years old...

The current Ansible Community Package includes 7.2.0 and 5.1.2, respectively. I see a problem there. And it looks like Ansible 12 will even include ansible.netcommon 8.0.0 or later.

felixfontein: This is a violation of the Ansible community package collection requirements: https://docs.ansible.com/ansible/devel/community/collection_contributors/collection_requirements.html#collection-dependencies

Never use `<` unless you have a **very** good reason for it (especially if it's not for a major release), and you make sure you quickly update it to new releases once they're out.

dupondje: Thanks for the useful comment.
Seems to be added here in the past: https://github.com/oVirt/ovirt-ansible-collection/pull/556

We will have a look and remove that versioned dependency.

PR #778, opened 2025-11-18
Rework hosted_engine_setup to support new ovirt-engine-appliance(s) built with kiwi

peter-boden:
Rework hosted_engine_setup to support new ovirt-engine-appliance(s) built with kiwi.
Added:
- devcontainer to easily run nox
- support for both ova and qcow2 appliances
- uefi support
- sanity ignore file for 2.21

Fixed a few small other issues


dupondje (changes requested):
2 things, otherwise :+1:

PR #787, opened 2026-04-08
ovirt_host_network: commit on ip/properties change

dupondje:
When 'save' is set to 'True', then we except that the changes are saved. Now this wasn't the case when for example only the IP address of an attachment was changed, but no new attachment was added.

This was because the `has_update` function does not return True when there was a change in update_custom_properties or update_address. Making it return True does not fix the issue neither, as oVirt does not save the changes in the 'setupnetworks' call when the IP was already changed before (via the update_address call).

Therefor, the fix here is to call commit_net_config when there was a change but setupnetworks was not executed.
If there for example is an IP change and a new attachment, then the 'commit_on_success' in 'setupnetworks' do cause everything to get saved, even if the IP was already modified in update_address.

### DIFF
--- a/changelogs/fragments/787-ovirt_host_network-commit_on_ip_change.yml
+++ b/changelogs/fragments/787-ovirt_host_network-commit_on_ip_change.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - ovirt_host_network - commit on ip/properties change (https://github.com/oVirt/ovirt-ansible-collection/pull/787).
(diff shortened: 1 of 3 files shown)


ferrebarrat (approved):

ferrebarrat (approved):

peter-boden (changes requested):
In the commit msgs:

> When 'save' is set to 'True', then we except...

expect

> Making it return True does not fix the issue neither

either

Otherwise lgtm.

PR #788, opened 2026-05-05
ovirt_snapshot: fail task when engine snapshot job fails (#777)

### LINKED ISSUE #777: Ovirt Snapshot successful when actual task had failed in ovirt
##### SUMMARY
A snapshot creation fails in ovirt but ansible module doesnt fail and provides exit code successful

##### COMPONENT NAME
ovirt.ovirt.ovirt_snapshot

##### STEPS TO REPRODUCE
Snapshot of vm was created with below 
```
ovirt_snapshot:
  vm_name: test
  use_memory: no
  wait: yes
  description: snapshotnew
```

In ovirt the snapshot task had failed. But ansible play was successful. 

##### EXPECTED RESULTS
In all cases if snapshot has failed in ovirt, ansible task should also fail

##### ACTUAL RESULTS

```

```

InfraArtists:
## Summary
Fixes #777 — `ovirt_snapshot` reported success even when the underlying snapshot creation job failed in oVirt engine.

The module now tags the snapshot creation call with a correlation ID, then queries the engine's jobs service after the wait completes and fails the task if the matching job ended in `FAILED`, `ABORTED`, or `UNKNOWN`. The wait condition is also hardened to handle the case where the engine cleans up the snapshot on job failure (entity becomes `None`), and the returned snapshot entity is refreshed so the reported status reflects the final state.

## Changes
- Generate a correlation ID and pass it to `snapshots_service.add(...)`.
- After waiting for the snapshot to reach `OK`, look up jobs by correlation ID and raise if any non-success status is found.
- Make the wait `condition` `None`-safe and add `fail_condition=lambda snap: snap is None`.
- Refresh the snapshot entity before returning so the result reflects the final engine state.
- Added changelog fragment.

### DIFF
--- a/plugins/modules/ovirt_snapshot.py
+++ b/plugins/modules/ovirt_snapshot.py
@@ -204,6 +204,7 @@ except ImportError:
 import os
 import ssl
 import time
+import uuid
 
 from ansible.module_utils.six.moves.http_client import HTTPSConnection
 from ansible.module_utils.six.moves.urllib.parse import urlparse
@@ -364,6 +365,7 @@ def create_snapshot(module, vm_service, snapshots_service, connection):
         snapshots_service.snapshot_service(module.params['snapshot_id'])
     )
     if snapshot is None:
+        correlation_id = str(uuid.uuid4())
         if not module.check_mode:
             disk_attachments_id = set(
                 get_disk_attachment(disk, vm_service.disk_attachments_service().list(), connection).id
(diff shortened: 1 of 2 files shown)


InfraArtists:
@dupondje Can you please also have a look of this PR? I add a good commit message : )

plugins/modules/ovirt_snapshot.py
```diff
@@ -375,15 +377,26 @@ def create_snapshot(module, vm_service, snapshots_service, connection):
...
                     persist_memorystate=module.params.get('use_memory'),
                     disk_attachments=[otypes.DiskAttachment(disk=otypes.Disk(id=da_id)) for da_id in disk_attachments_id] if disk_attachments_id else None
-                )
+                ),
+                query={'correlation_id': correlation_id},
             )
         changed = True
+        snapshot_service = snapshots_service.snapshot_service(snapshot.id)
         wait(
-            service=snapshots_service.snapshot_service(snapshot.id),
```
dupondje, line 382: On first sight it seems like the existing code should already work?
It poll's the snapshot status via the snapshot_service for the just created snapshot.
And if wait is true, it waits until timeout that snapshot is ok.

Did you debug why that didn't work as expected?

InfraArtists: You're right. the existing wait already covers the cases where the snapshot is deleted, stuck locked, or ends in a non-OK status.
The remaining gap is when the snapshot resource itself reaches OK but an associated engine job (memory-state save, post-create sub-task, etc.) ends in FAILED/ABORTED.
The snapshot entity status doesn't reflect the full command's outcome, only the engine job does.

I haven't reproduced the original report end-to-end, so I can't confirm that's the exact code path the reporter hit. Before merging, it'd be worth asking @hemak88 for:

the final snapshot_status after the (apparent) success
the engine job status of engine.log excerpt around the failure

If they hit one of the cases the existing wait already covers, the right fix is probably just to make the None case raise a clearer error instead of an AttributeError rather than the broader correlation-ID approach.

dupondje: Please stop using AI agents ...
If you don't even understand the code yourself, this is useless.

InfraArtists: @dupondje  I'm sorry if I bothered you. You are right, and it's my mistake.
It's better to focus on the Ansible side.
Sorry again.

dupondje:
Ai bot out of credits. Closing.
Feel free to reopen once you fully understand what's happening and that root cause has been fixed :)

PR #792, opened 2026-05-08
Fix dmsetup deps failure on failed multipath devices

### LINKED ISSUE #791: Fix dmsetup deps failure on failed multipath devices in remove_mpath_device.yml
Add error handling to continue when dmsetup deps fails on devices that are already in a failed state or removed from device-mapper. This prevents the role from failing when encountering devices it's specifically meant to clean up.

kelwhite:
Fixes #791

### DIFF
--- a/roles/remove_stale_lun/tasks/remove_mpath_device.yml
+++ b/roles/remove_stale_lun/tasks/remove_mpath_device.yml
@@ -6,7 +6,7 @@
 - name: Get underlying disks (paths) for a multipath device(s) and turn them into a list.
   ansible.builtin.shell: >-
     for dev in {{ lun_wwid }}; do
-      set -euo pipefail && dmsetup deps -o devname $dev | cut -f 2 | cut -c 3- | tr -d "()" | tr "\r\n" " ";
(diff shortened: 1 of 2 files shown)


roles/remove_stale_lun/tasks/remove_mpath_device.yml
```diff
@@ -6,7 +6,7 @@
 - name: Get underlying disks (paths) for a multipath device(s) and turn them into a list.
   ansible.builtin.shell: >-
     for dev in {{ lun_wwid }}; do
-      set -euo pipefail && dmsetup deps -o devname $dev | cut -f 2 | cut -c 3- | tr -d "()" | tr "\r\n" " ";
+      set -euo pipefail && dmsetup deps -o devname $dev 2>/dev/null | cut -f 2 | cut -c 3- | tr -d "()" | tr "\r\n" " ";
```
dupondje, line 9: Have you tested this?
Cause afaik this fix is not valid. 

Why? Well, we set `set -euo pipefail` first. Which causes command to fail as a whole if one command in the row returns a non-zero exit code. I guess dmsetup deps will still return a non-zero exit code when there are failed multipath devices.
The result of your change is that the error will not be visible anymore, but the command will still fail.

dupondje (changes requested):

dupondje:
@kelwhite : please reopen when feedback/fix.

PR #793, opened 2026-06-19
Feature/nvme overtcp

tdemaret:
Add NVMe-oF/TCP storage domain support to oVirt. This is a block/shared storage domain type, analogous to iSCSI, using the built-in Linux NVMe 
initiator (nvme-cli). It supports both authenticated (DH-HMAC-CHAP) and non-authenticated NVMe-oF targets.

The feature try to be backend-agnostic for any NVMe/TCP target works (Ceph NVMe-oF Gateway is the reference implementation).

# Change :
- New storage domain type: nvmeof in REST API / Ansible / SDK
- Authentication : optional DH-HMAC-CHAP 
- Multipath : native NVMe multipath detected automatically; falls back to dm-mpath
- Host deployment: NVMe host NQN auto-detected from /etc/nvme/hostnqn

# Limitations :
- No iSCSI bond support for NVMe-oF
- NVMe-oF storage domain cannot be edited or imported via webadmin


plugins/modules/ovirt_host.py
```diff
@@ -492,50 +540,61 @@ def main():
...
-        iscsi=dict(default=None, type='dict'),
-        check_upgrade=dict(default=True, type='bool'),
-        reboot_after_upgrade=dict(default=True, type='bool'),
-        vgpu_placement=dict(default=None, choices=['consolidated', 'separated']),
+        kernel_params=dict(default=None, type="list", elements="str"),
+        hosted_engine=dict(default=None, choices=["deploy", "undeploy"]),
+        power_management_enabled=dict(default=None, type="bool"),
+        activate=dict(default=True, type="bool"),
+        iscsi=dict(default=None, type="dict"),
+        nvmeof=dict(default=None, type="dict"),
```
peter-boden, line 579: This will leak `dhchap_key` if ansible is used with more verbosity. (use `no_log` like it used in e.g. `ovirt_host_storage_info.py`)

plugins/modules/ovirt_storage_connection.py
```diff
@@ -211,6 +216,7 @@ def update_check(self, entity):
             equal(self.param('username'), entity.username) and
             equal(self.param('port'), entity.port) and
             equal(self.param('target'), entity.target) and
+            equal(self.param('nqn'), entity.nqn) and
```
peter-boden, line 219: This module exposes `nqn` but not `transport`/`trsvcid`/`host_nqn`/`dhchap_key` — so a full NVMe-oF storage connection can only be created through `ovirt_host: state: nvmeofconnect`, not through this general-purpose storage-connection module.

Is that intentional (NVMe connections being host-scoped rather than pool-scoped like iSCSI/NFS), or should this module gain the remaining fields for parity?

roles/host_nvmeof_setup/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
+---
```
peter-boden, line 1: This whole file is basically wrong. 
See my comments in the [ovirt-engine PR](https://github.com/oVirt/ovirt-engine/pull/1169/changes#diff-66280480a5ba08ad82788de0965a0b8bb9f79d00b26c04b70d0e49f2c17a09d3)

peter-boden (changes requested):

PR #795, opened 2026-07-10
build: update build flow

dupondje:
- Update the way version tags are added during build. Now we can tag a commit in github, and the correct version string will be used for the builds.
- Run the workflow on any tag, so it builds tags in branches also.
- Drop copr


.github/workflows/build.yaml
```diff
@@ -42,28 +44,64 @@ jobs:
...
+            shortcut: el10s
             container-name: el10stream
     container:
       image: quay.io/ovirt/buildcontainer:${{ matrix.container-name }}
+
     steps:
-      - name: Install python3 modules
-        run: ${{ matrix.pip-command }} install pycodestyle pylint voluptuous yamllint rstcheck antsibull-changelog rich ansible-lint cryptography
+      - name: Checkout
+        uses: actions/checkout@v7
```
peter-boden, line 64: Switching away from `ovirt/checkout-action` breaks the changelog check in `.automation/build.sh`.
[Fetch depth is 0](https://github.com/oVirt/checkout-action/blob/main/action.yml#L6) for the ovirt action, while it is [1 for the github action](https://github.com/actions/checkout/blob/main/action.yml#L76).

[build.sh calls a diff with ^HEAD](https://github.com/oVirt/ovirt-ansible-collection/blob/master/.automation/build.sh#L58-L61), this fails with fetch depth 1:
```
➜ git clone https://github.com/oVirt/ovirt-ansible-collection --depth=1
Cloning into 'ovirt-ansible-collection'...
remote: Enumerating objects: 503, done.
remote: Counting objects: 100% (503/503), done.
remote: Compressing objects: 100% (436/436), done.
remote: Total 503 (delta 78), reused 223 (delta 21), pack-reused 0 (from 0)
Receiving objects: 100% (503/503), 398.39 KiB | 3.19 MiB/s, done.
Resolving deltas: 100% (78/78), done.
➜ cd ovirt-ansible-collection
➜ git diff --quiet HEAD^ ./plugins ./roles 2>&1; echo "exit=$?"
fatal: ambiguous argument 'HEAD^': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
exit=128
```

peter-boden (changes requested):

.github/workflows/build.yaml
dupondje: Fixed by adding fetch-depth: 2

peter-boden (approved):

